PR #1, opened 2022-07-24
feat(chungthuang): Use sqlite for local development

chungthuang:
This removes the need to run postgres with docker

### DIFF
--- a/config/env/development/database.js
+++ b/config/env/development/database.js
@@ -1,0 +1,10 @@
+module.exports = ({ env }) => ({
+  connection: {
+    client: 'sqlite',
+    connection: {
+      filename: env('DATABASE_FILENAME', '.tmp/data.db'),
+      schema: false,
+    },
+    useNullAsDefault: true,
+  },
+});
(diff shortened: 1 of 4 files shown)


config/env/development/database.js
```diff
@@ -0,0 +1,10 @@
+module.exports = ({ env }) => ({
+  connection: {
+    client: 'sqlite',
+    connection: {
+      filename: env('DATABASE_FILENAME', '.tmp/data.db'),
+      schema: false,
+    },
+    useNullAsDefault: true,
+  },
```
jiawen195, line 9: What is this section of code doing?

chungthuang: This is exporting an object to configure the database used by strapi
```
{
  connection: {
    client: 'sqlite',
    connection: {
      filename: env('DATABASE_FILENAME', '.tmp/data.db'),
      schema: false,
    },
    useNullAsDefault: true,
  },
}
```
It says use the sqlite client to connect to the database. The sqlite server writes to a file defined by the environment variable `DATABASE_FILENAME`. If the env var is not provided, use `.tmp/data.db`.

PR #7, opened 2022-08-15
fix(chungthuang): Make hero image optional in project collection

chungthuang:
This makes it easier to create projects in development and testing


Enjoy2Live:
Why don't we make it required only in prod?

Currently the UI would look off without a heroImage but maybe I can fallback to a default image if nothing is specified.

chungthuang:
The schema is a json file, not code, so I can't easily change it based on the environment. Can we go with a default image?

PR #15, opened 2023-02-20
Dbradham/add likes

dbradham:
I added a new collection for likes to be used in a generic way. (For ideas, comments, and whatever else users want to "like")

I also added a .gitignore file to make it easier to commit changes.

The package-json was changed when I ran the app and I committed the change because the previous version said the app name was "my-project" and now it says "devlaunchers/strapiv4" (this change happened automatically when the app ran)


src/api/like/content-types/like/schema.json
```diff
@@ -0,0 +1,25 @@
...
+  "pluginOptions": {},
+  "attributes": {
+    "objectId": {
+      "type": "string"
+    },
+    "objectType": {
+      "type": "string"
+    },
+    "userId": {
+      "type": "string"
```
chungthuang, line 22: We can use a relation field linking to the users collection, like https://github.com/dev-launchers/strapiv4/blob/main/src/api/idea-card/content-types/idea-card/schema.json#L53.

src/api/like/content-types/like/schema.json
```diff
@@ -0,0 +1,25 @@
...
+  "options": {
+    "draftAndPublish": true
+  },
+  "pluginOptions": {},
+  "attributes": {
+    "objectId": {
+      "type": "string"
+    },
+    "objectType": {
+      "type": "string"
```
chungthuang, line 19: This is ideas, comments, etc? Maybe it can be modeled as an enum.

src/api/like/content-types/like/schema.json
```diff
@@ -0,0 +1,25 @@
...
+    "displayName": "Like",
+    "description": ""
+  },
+  "options": {
+    "draftAndPublish": true
+  },
+  "pluginOptions": {},
+  "attributes": {
+    "objectId": {
+      "type": "string"
```
chungthuang, line 16: This is a UID if it's pointing to other collections.

src/api/like/content-types/like/schema.json
dbradham: Thanks for the feedback and examples here. I have addressed all 3 with my latest commit @chungthuang

chungthuang (approved):

PR #16, opened 2023-02-22
Google meet test

oriyomibadmus:


src/api/google-meet/content-types/google-meet/lifecycle.js
```diff
@@ -0,0 +1,12 @@
+'use strict';
+const { sanitizeEntity } = require('@strapi/utils');
+
+module.exports = {
+  async afterFindOne(ctx) {
+    const { name } = ctx.params;
+    const meet = await strapi.services['google-meets'].findOne({ name: name });
```
chungthuang, line 7: Would this lead to infinite loop? I think we can get the meet from ctx.result?

src/api/google-meet/controllers/google-meet.js
```diff
@@ -6,4 +6,12 @@
 
 const { createCoreController } = require('@strapi/strapi').factories;
 
-module.exports = createCoreController('api::google-meet.google-meet');
+module.exports = createCoreController('api::google-meet.google-meet', ({strapi}) => ({
+    userFacingMeets(internalMeet) {
```
chungthuang, line 10: Where is this used?

src/api/google-meet/content-types/google-meet/lifecycle.js
chungthuang: Oh this is old file that we should delete.

PR #22, opened 2023-03-19
Gtoutin/new applicant discord notifier

gtoutin:
- Dev Launchers bot pushes to #api-gateway whenever a new applicant is created
- docker dev env improvements (now it reloads via a bind mount)


Dockerfile
```diff
@@ -9,4 +9,4 @@ COPY public ./public
 COPY src ./src
 RUN npm run build
 EXPOSE 1337
-CMD ["npm", "run", "start"]
+CMD ["npm", "run", "develop"]
```
chungthuang, line 12: This is also used to build the staging image in https://github.com/dev-launchers/strapiv4/blob/main/.github/workflows/staging-image.yaml#L47. Can we override the entrypoint in https://github.com/dev-launchers/strapiv4/blob/main/docker-run.sh#L2 instead

Makefile
```diff
@@ -0,0 +1,8 @@
+build-docker:
+	docker build -t strapiv4 .
+
+run-docker:
```
chungthuang, line 4: Could we remove docker-run.sh, or call docker-run.sh here? They seem to be doing the same thing.

Makefile
chungthuang, line 1: Can you add how to use this in README?

src/api/applicant/content-types/applicant/lifecycles.js
```diff
@@ -1,14 +1,54 @@
+const _ = require('lodash');
```
chungthuang, line 1: How are we using loadash?

src/api/applicant/content-types/applicant/lifecycles.js
```diff
@@ -1,14 +1,54 @@
...
+// create new Webhook
+const hook = new Webhook(
+  `${process.env.OPEN_POSITIONS_DISCORD_WEBHOOK}`
+);
+
+hook.setUsername("New Applicant Notifier")
+.setAvatar('https://avatars.githubusercontent.com/u/53379976?s=200&v=4');
+
+const formatJSON = (jsonData) => {
+  return `\`\`\`JSON\n${JSON.stringify(jsonData, null, 1)}\n\`\`\``;
```
chungthuang, line 14: What are all the escaped slashes for?

chungthuang:
This is looking good! Is it possible to make this into a reusable service, so we can easily send notifications in other collections.

src/api/applicant/content-types/applicant/lifecycles.js
```diff
@@ -1,14 +1,54 @@
...
+      where: { id: project }
+    });
+    try {
+      // insert slug and title into the applicant project obj
+      const projectSlug = applicantProject.slug;
+      event.model.attributes.project.slug = projectSlug;
+      event.model.attributes.project.title = applicantProject.title;
+    } 
+    catch (e) {
+      console.log("couldn't find project slug");
```
chungthuang, line 34: I think it'll be helpful to log the project ID/name.

src/api/applicant/content-types/applicant/lifecycles.js
```diff
@@ -1,14 +1,54 @@
...
-    async beforeCreate(event) {
+  async beforeCreate(event) {
+    // get the project associated with this applicant from the db
     const project = event.params.data.project;
     const applicantProject = await strapi.db.query('api::project.project').findOne({
-        where: {slug: project}
-      });
-    const projectID = applicantProject.id;
-    event.params.data.project = projectID
+      where: { id: project }
```
chungthuang, line 25: I thought the frontend only passes the project name.

src/api/applicant/content-types/applicant/lifecycles.js
gtoutin: We're using it here https://github.com/dev-launchers/strapiv4/pull/22/files#diff-6144805fbeed9ef42cece8f4c59076e25b68343b142ce530301ffd522cc6add2R41

`var relevantInfo = _.omit(event.params.data, ['email', 'age', 'zip', 'createdBy', 'updatedBy', 'createdAt', 'updatedAt', 'publishedAt']);`

It's a bit confusing with the underscore, I'll spell out `lodash` instead for readability

.env.example
```diff
@@ -1,4 +1,6 @@
+# IMPORTANT: do not put quotes "" in the URL variables
```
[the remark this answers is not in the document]
gtoutin: If there are quotes, it will complain about the URL not being in the correct format. It thinks the quotes are actually part of the URL.
Maybe I could add some string sanitizing before the vars get used so it can be more robust.

Makefile
gtoutin: I split it out because if the image goes down, I don't want to do the building step again. Now that the dev dir is mounted, just running the container again is enough to get it back up with changes that were not there when the container was built.

Makefile
gtoutin: Sure!

src/api/applicant/content-types/applicant/lifecycles.js
gtoutin: I think they're there so that the backticks can show up.
I think I could replace it with
```suggestion
  return `"```JSON\n" + `${JSON.stringify(jsonData, null, 1)}` + "\n```"`;
```

Makefile
chungthuang: Oh I see. Maybe we can just remove docker-run.sh and document how to use the make commands in the README.

src/api/applicant/content-types/applicant/lifecycles.js
oriyomibadmus: I think there's might be a difference in the data being returned when a new applicant is created using the Strapi UI and the data returned when a new applicant is created using the POST request with the structure of the "request body" gotten from the frontend

Dockerfile
gtoutin: I added a separate dev dockerfile and used it to build the dev docker image in the makefile

src/api/applicant/content-types/applicant/lifecycles.js
```diff
@@ -1,14 +1,62 @@
...
+    } 
+    catch (e) {
+      console.log(`couldn't find project slug for project ${project}`);
+      event.model.attributes.project.slug = "";
+    }
+  },
+  async afterCreate(event) {
+    // insert proj title and remove unnecessary info from message
+    event.params.data.project = event.model.attributes.project.title;
+    var relevantInfo = lodash.omit(event.params.data, ['email', 'age', 'zip', 'createdBy', 'updatedBy', 'createdAt', 'updatedAt', 'publishedAt']);
```
chungthuang, line 49: What if we add more fields to the projects collection? Maybe we can select the needed fields instead of omitting the the unwanted ones?

src/api/applicant/content-types/applicant/lifecycles.js
chungthuang: Bump this comment. I think we still need to query the project by name(slug).

oriyomibadmus (approved):

Miminiverse (approved):
I have tested it and it is working great!

src/api/applicant/content-types/applicant/lifecycles.js
```diff
@@ -1,14 +1,76 @@
...
+    const newApplicantMsg = new MessageBuilder()
+      .setTitle("New Applicant!")
+      .setURL(`${frontend_url_sanitized}/projects/${event.model.attributes.project.slug}`)
+      .setDescription(
+        formatJSON(relevantInfo)
+      );
+
+    // create new Webhook
+    const hook = new Webhook(
+      event.model.attributes.project.discordWebhookUrl
```
chungthuang, line 68: Note to myself. Not all projects will set up a discord webhook URL. Need to add a check here.

src/api/applicant/content-types/applicant/lifecycles.js
```diff
@@ -1,14 +1,76 @@
...
+    );
+
+    hook.setUsername("New Applicant Notifier")
+      .setAvatar('https://avatars.githubusercontent.com/u/53379976?s=200&v=4');
```
dapowers87, line 72: Can we have this URL exported to a common location as its own variable?

src/api/applicant/content-types/applicant/lifecycles.js
```diff
@@ -1,14 +1,76 @@
+const lodash = require('lodash');
+const fetch = require('node-fetch');
 const { createModuleResolutionCache } = require("typescript");
+const { Webhook, MessageBuilder } = require('discord-webhook-node');
+
+const frontend_url_sanitized = process.env.FRONTEND_URL.replace(/["']/g, "");
+
+// use markdown block code to format this as JSON
+const formatJSON = (jsonData) => {
+  return "```JSON\n" + JSON.stringify(jsonData, null, 1) + "\n```";
```
dapowers87, line 10: Consider using an npm package like `@braintree/sanitize-url` and use its more comprehensive `sanitizeUrl` method

chungthuang: Actually this is not to sanitize URL but to format a string as JSON markdown. Do you have suggestions to better implement this?

PR #33, opened 2023-04-07
Added user recommendations with caching

nmoon4:
Sending a GET request to /users/user-recommendations will return 6 recommended users based on common interests and their similarity scores. Caching will store the 100 random users for 60 sec.


src/api/user-recommendations/controllers/user-recommendations.js
```diff
@@ -0,0 +1,81 @@
...
+    // check if cache already has random users
+    if (users == undefined) {
+      const userCount = await strapi.entityService.count(
+        'plugin::users-permissions.user'
+      );
+
+      // gets 100 random user ids
+      const randomUserIds = Array(100)
+        .fill()
+        .map(() => Math.floor(Math.random() * userCount) + 1);
```
chungthuang, line 25: How do we make sure there is no duplicate?

src/api/user-recommendations/controllers/user-recommendations.js
```diff
@@ -0,0 +1,81 @@
...
+  async recommend(ctx) {
+    let users = myCache.get('randomUsers');
+
+    // check if cache already has random users
+    if (users == undefined) {
+      const userCount = await strapi.entityService.count(
+        'plugin::users-permissions.user'
+      );
+
+      // gets 100 random user ids
```
chungthuang, line 22: There are less than 100 users in staging.

src/api/user-recommendations/controllers/user-recommendations.js
```diff
@@ -0,0 +1,81 @@
...
+/**
+ * Read the documentation (https://strapi.io/documentation/developer-docs/latest/development/backend-customization.html#core-controllers)
+ * to customize this controller
+ */
+
+const myCache = new NodeCache({ stdTTL: 60 });
+
+module.exports = {
+  async recommend(ctx) {
+    let users = myCache.get('randomUsers');
```
chungthuang, line 14: So this cache will only have 1 key?

src/api/user-recommendations/controllers/user-recommendations.js
```diff
@@ -0,0 +1,81 @@
...
+          }
+        }
+
+        userScores.set(user.id, numMatches);
+      }
+    }
+
+    // sort greatest to least based on similarity score
+    users.sort((a, b) => userScores.get(b.id) - userScores.get(a.id));
+    users = users.slice(0, 6);
```
chungthuang, line 71: Let's make 6 a const.

src/api/user-recommendations/controllers/user-recommendations.js
nmoon4: I tested it on my local environment with 150 users in the database and usually ended up getting about 70 unique random ids. If there are duplicates, then less than 100 random users will be selected, but recommendations can still be made. I could change it so that there will always be 100 unique random ids, if that would be better.

src/api/user-recommendations/controllers/user-recommendations.js
nmoon4: Yes, having all of the random users under one key makes it easy to access all of them at the same time. Alternatively I could have a key for each random user if that would better.

src/api/user-recommendations/controllers/user-recommendations.js
nmoon4: Could I add more users to staging? I was able to quickly make 150 users on my local environment.

chungthuang: In staging we use Google for authentication, so each user needs a Google account.

src/api/user-recommendations/controllers/user-recommendations.js
chungthuang: I think that is a product question. I'm okay with reusing the same users if the request comes within the same minute (TTL of cache).

src/api/user-recommendations/controllers/user-recommendations.js
chungthuang: I'm okay with this. As we get more users, the probability of duplication gets lower.

src/api/user-recommendations/controllers/user-recommendations.js
nmoon4: I think as the code is currently, it would still be able to provide recommendations even if there's less than 100 total users. The only issue is that there wouldn't be any randomness. Another way to test in staging would be if instead of 100 random users, I grab about half of the total users and rank them by similarity.

chungthuang (approved):

PR #42, opened 2023-05-16
feat(chungthuang): Install plugin to return nested relations

chungthuang:
By default, strapi wouldn't populate relational fields. The frontend can request it by adding `?populate=*`, but this only populate one-level deep. For projects collections, the team field is 2 level-deep. The query string becomes hard to craft, so we use this plugin that will return deeply nested relations.


Enjoy2Live (approved):

VictorDiniz89 (commented):
LGTM

PR #45, opened 2023-05-23
Send errors encountered

VictorDiniz89:


config/plugins.js
```diff
@@ -13,7 +13,7 @@ module.exports = ({env}) => ({
     sentry: {
       enabled: true,
```
chungthuang, line 14: Can we enable this based on environment? We only need to collect from staging and production.

chungthuang: Oh never mind, development and testing uses the config in https://github.com/dev-launchers/strapiv4/tree/main/config/env.

config/plugins.js
```diff
@@ -13,7 +13,7 @@ module.exports = ({env}) => ({
...
       config: {
-        dsn: 'https://1b4a30f38ea04b9b956481f226a83e4e@o1123967.ingest.sentry.io/6372271',
+        dsn: 'https://9b74d2bd2a50432da68982c3c4d72420@o4505235023921152.ingest.sentry.io/4505235032899584',
```
chungthuang, line 16: Can we read is from environment? That way we can inject the dsn as a secret. See `jwtSecret` for an example.

chungthuang:
Can you update the commit message to follow https://github.com/semantic-release/semantic-release#commit-message-format?

chungthuang:
Foll

> Can you update the commit message to follow https://github.com/semantic-release/semantic-release#commit-message-format?

You can update commit message when you squashed multiple commits.

PR #53, opened 2023-06-19
Added completedOnboarding variable to user model

nmoon4:
The completeOnboarding variable will keep track of whether the user has completed the onboarding process.


src/extensions/users-permissions/content-types/user/schema.json
```diff
@@ -150,6 +150,10 @@
     },
     "schedule": {
       "type": "json"
+    },
+    "completedOnboarding": {
```
chungthuang, line 154: How would this be set?

chungthuang (approved):

PR #90, opened 2023-08-27
fix(cthuang): Redirect user to signup or user profile page after sign in

### LINKED ISSUE #89: [Bug] User not redirected after authentication
**Describe the bug**
After user go through the Oauth flow to sign in, they got the token but not redirected back to the user profile page.
image

**To Reproduce**
1. Go to https://staging.devlaunchers.org/
2. Click Sign In

**Expected behavior**
User is redirected to the staging website.

**Additional context**[
@dbradham noticed https://github.com/dev-launchers/strapi/blob/main/extensions/users-permissions/controllers/Auth.js#L219-L225 is not migrated to v4.

chungthuang:
fixes #89

### DIFF
--- a/src/extensions/users-permissions/server/middlewares/jwt.js
+++ b/src/extensions/users-permissions/server/middlewares/jwt.js
@@ -13,7 +13,11 @@ module.exports = (options, { strapi }) => {
     // Check if the route is local login or login with a provider
     // Set token in cookie
     if (ctx.url.startsWith('/api/auth/')) {
-      const token = ctx.response.body.jwt;
+      const body = ctx.response.body;
+      if (!body) {
+        return
+      }
+      const token = body.jwt;
       if (token) {
         /*
           if in dev env, disable secure option because secure option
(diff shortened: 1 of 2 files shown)


dbradham (commented):
Looks good to me but I was getting some strange behavior testing this locally which is different than what I've seen on staging... I was able to select my google account as expected, and then a request was sent to my strapi local's relative path at: `/api/connect/google/api/auth/google/callback?id_token=ey` instead of `/api/auth/...` so the conditional to set the cookie wasn't even executing. Likely related to something in my local config.

dbradham (approved):

chungthuang:
> Looks good to me but I was getting some strange behavior testing this locally which is different than what I've seen on staging... I was able to select my google account as expected, and then a request was sent to my strapi local's relative path at: `/api/connect/google/api/auth/google/callback?id_token=ey` instead of `/api/auth/...` so the conditional to set the cookie wasn't even executing. Likely related to something in my local config.

In http://localhost:1337/admin/settings/users-permissions/providers, what is `The redirect URL to your front-end app`? I put `http://localhost:1337/api/auth/google/callback` and pass the actual frontend URL in the `redirectURL` query parameter, so the request looks like http://localhost:1337/api/connect/google?redirectURL=http://localhost:3000/users/me

src/extensions/users-permissions/server/middlewares/jwt.js
```diff
@@ -32,6 +32,21 @@ module.exports = (options, { strapi }) => {
             sameSite: 'None'
           });
         }
+
+        const grant = ctx.session.grant;
```
chungthuang, line 36: @dapowers87 any suggestion to make this easier to read?

Issue #35, opened 2023-04-10
Allow users to update their profile picture


chungthuang:
We are using Azure as our storage provider. We can upload images through the admin dashboard. Now we need an API so the User Profile team can build a feature that allow users to change their profile picture.

PR #93, opened 2023-09-04
feat(oriyomi): Allow users to update their profile picture #35

oriyomibadmus:


config/middlewares.js
```diff
@@ -22,7 +22,44 @@ const DevLaunchersSubdomains = [
...
   'strapi::errors',
-  'strapi::security',
+
+  {
+    name: "strapi::security",
+    config: {
+      contentSecurityPolicy: {
+        useDefaults: true,
+        directives: {
+          "connect-src": ["'self'", "https:"],
```
chungthuang, line 32: Where did you get these values from? Is there any documentation that you can link to?

config/middlewares.js
Miminiverse, line 1: Can you let me know how can I get the .env info in here?
process.env.STORAGE_URL
        account: env("STORAGE_ACCOUNT"),
        accountKey: env("STORAGE_ACCOUNT_KEY"),
        serviceBaseURL: env("STORAGE_URL"), // optional
        containerName: env("STORAGE_CONTAINER_NAME"),

oriyomibadmus: Reach out to @chungthuang for the Azure credentials

src/api/profile/content-types/profile/lifecycles.js
Miminiverse, line 1: Should we add a condition to handle cases when there's no event.params.data.profilePictureUrl or event.result.profilePicture? When I save a user without putting the URL the app will break, while I believe these fields could be null

chungthuang: That's a great suggestion! In general, we should make sure parameters are present before using them.

oriyomibadmus: Nice! Currently, users are registered through Google auth, which means the profile URL is auto-completed. However, as you suggested, I believe the condition needs to be handled to avoid breaking. Thanks!

chungthuang (approved):

PR #95, opened 2023-09-08
Implement "Save Idea" Functionality #913

Rutuja177:
Implement "Save Idea" Functionality

**Description**:
This pull request introduces the "Save Idea" functionality to allow users to save ideas. The key changes and additions are as follows:

1. Created a "SavedIdea" collection to represent the relationship between users and saved ideas.
2. Added relation fields for "users" (many-to-many) and "Idea_card" (one-to-one) in the "SavedIdeas" collection.
3. Included a "SavedAt" timestamp to track when an idea was saved.

Thank you!


src/api/save-idea/content-types/save-idea/schema.json
```diff
@@ -0,0 +1,33 @@
...
+    "draftAndPublish": true
+  },
+  "pluginOptions": {},
+  "attributes": {
+    "ObjectId": {
+      "type": "uid"
+    },
+    "users": {
+      "type": "relation",
+      "relation": "manyToMany",
```
dbradham, line 20: Should this be oneToOne instead?

src/api/save-idea/content-types/save-idea/schema.json
```diff
@@ -0,0 +1,33 @@
...
+      "target": "plugin::users-permissions.user",
+      "inversedBy": "save_ideas"
+    },
+    "SavedAt": {
+      "type": "datetime"
+    },
+    "idea_card": {
```
dbradham, line 27: if we already have the `ObjectId`, is this necessary?

Rutuja177: The updates have been implemented.

src/api/save-idea/content-types/save-idea/schema.json
Rutuja177: The updates have been implemented.

dbradham (approved):
Looks good!

Miminiverse:
Can we be more explicit in the collection's displayName instead of Save? Or the name is used for a certain purpose?

chungthuang (approved):

PR #99, opened 2023-09-13
Adding Project Resources API

yacineKahlerras:
- created schema and for [Project Resources](https://projects.devlaunchers.org/projects/resources) page
- added controller to project-resources to filter fields out


src/api/project-resource/controllers/project-resource.js
```diff
@@ -0,0 +1,73 @@
+"use strict";
+
+/**
+ * project-resource controller
+ */
+
+const { createCoreController } = require("@strapi/strapi").factories;
+
+module.exports = createCoreController(
```
chungthuang, line 9: The pattern we've been using is default controller with lifecycle hook for customization. For example, the applicant collection uses strapi's default to create the CRUD endpoints. controller https://github.com/dev-launchers/strapiv4/blob/main/src/api/applicant/routes/applicant.js. Then we use https://github.com/dev-launchers/strapiv4/blob/main/src/api/applicant/content-types/applicant/lifecycles.js to customize the create method.

chungthuang: @oriyomibadmus has a lot of experience using lifecycles.

yacineKahlerras: > @oriyomibadmus has a lot of experience using lifecycles.

ah okay i've used the same pattern i saw in the project api, i'll make some changes and ask yomi to take a look

PR #100, opened 2023-09-16
feat(miminiverse): add database

Miminiverse:


Miminiverse:
This is for testing adding data in sqlite

PR #101, opened 2023-09-16
feat(miminiverse): add postman test for project

Miminiverse:
This PR contains POSTMAN testing and  JSON schema validation for the endpoint /projects to ensure the format we receive is the same as expected.


integration-test/project.json
chungthuang, line 1: Should we merge this into https://github.com/dev-launchers/strapiv4/blob/main/integration-test/collection.json, so it's used by the integration test github action https://github.com/dev-launchers/strapiv4/blob/main/.github/workflows/test.yaml#L26?

integration-test/project.json
```diff
@@ -0,0 +1,170 @@
...
+					"script": {
+						"exec": [
+							"",
+							"pm.test(\"Status code is 200\", function () {",
+							"    pm.response.to.have.status(200);",
+							"});",
+							"",
+							"const project = pm.response.json();",
+							"",
+							"pm.test(\"Project data is valid\", () => {",
```
chungthuang, line 71: This helps with readability of the test!

integration-test/project.json
```diff
@@ -0,0 +1,170 @@
...
+					"path": [
+						"auth",
+						"local"
+					]
+				}
+			},
+			"response": []
+		},
+		{
+			"name": "get projects",
```
chungthuang, line 58: Nit: I would call this list projects, because the endpoint return multiple projects. `get project` would be something like `/projects/1`

integration-test/project.json
Miminiverse: Can we add more files in the integration test github action? For example ` newman run integration-test/project.json`
  I separate them because I'm going to create tests for other endpoints as well and I'm afraid it might be difficult to review and maintain if we put them all in one collection.json. If this collection.json is meant to contain all of the tests then I will merge them. Please let me know.

chungthuang: We can run `newman` multiple times. The only problem I see with multiple collections is we need to repeat some requests, like login.

chungthuang: It doesn't look like Postman support sharing common requests between collections yet https://github.com/postmanlabs/postman-app-support/issues/1535. But I see the appeal of having separate collection for each content type. It's easier to read and other project teams using the backend can just focus on the collection for their content type.

chungthuang (approved):

PR #103, opened 2023-11-28
feat(oriyomi): Create The Notification Collection and add hooks to ideaCard and Comment Collection

oriyomibadmus:


src/api/comment/content-types/comment/lifecycles.js
```diff
@@ -0,0 +1,23 @@
+module.exports = {
+
+    async afterCreate(event) {
+
+        const ideaCard = event.result.idea_card
+        const ideaName = event.result.idea_card.ideaName
+        const commentAuthor = event.result.author
+        const commentText = event.result.text
+        const timeCreated = event.result.createdAt
```
Enjoy2Live, line 9: maybe if we use object destructing here it'd be better

src/api/comment/content-types/comment/lifecycles.js
```diff
@@ -0,0 +1,23 @@
...
+
+        await strapi.entityService.create('api::notification.notification', {
+            data: {
+              Title: commentAuthor + " commented on " + ideaName,
+              Content: commentText,
+              Tag: "New Comment",
+              idea_card: ideaCard.id,
+              TimeCreated: timeCreated,
+              Read: false,
```
Enjoy2Live, line 18: how does the state changes from unread to read? are we supposed to send some sort of request to backend?

src/api/idea-card/content-types/idea-card/lifecycles.js
```diff
@@ -0,0 +1,23 @@
+module.exports = {
+
+    async afterCreate(event) {
+
+        const ideaCard = event.result.id
+        const ideaName = event.result.ideaName
+        const timeCreated = event.result.createdAt
```
Enjoy2Live, line 7: maybe if we use object destructing here it'd be better

Enjoy2Live (changes requested):
That's promising, thanks for implementing such a needed feature @oriyomibadmus .
I appropriate the simplicity of the PR and how everything is documenting itself.

I left some comments here to get better idea with some requested changes

src/api/notification/content-types/notification/schema.json
```diff
@@ -0,0 +1,36 @@
...
+      "relation": "oneToOne",
+      "target": "api::idea-card.idea-card"
+    },
+    "Content": {
+      "type": "string"
+    },
+    "Read": {
+      "type": "boolean"
+    },
+    "Tag": {
```
chungthuang, line 29: What does tag mean here?

src/api/notification/content-types/notification/schema.json
```diff
@@ -0,0 +1,36 @@
...
+  },
+  "options": {
+    "draftAndPublish": false
+  },
+  "pluginOptions": {},
+  "attributes": {
+    "Title": {
+      "type": "string"
+    },
+    "idea_card": {
```
chungthuang, line 18: Would we need to create a field for each new collection that wants notifications? Maybe we can create an enum field for the collection type (idea-card, comment, etc) and another field for ID of the object that triggered the notification.
The downside is we don't have a relationship at the database level, so we can't do things like cascade delete directly from database.

src/api/comment/content-types/comment/lifecycles.js
Enjoy2Live, line 18: how can we bind the notification to its appropriate user?
for example only idea_author and people interested in the idea would want to get notified but not everyone else

src/api/notification/content-types/notification/schema.json
chungthuang, line 1: Should we store the ID of the user that triggered the notification? For example, the user that commented on an idea.

chungthuang: Actually maybe we don't need it. We can get the author from the comment and author from the idea card.

src/api/comment/content-types/comment/lifecycles.js
chungthuang: Like `const {ideaCard, timeCreated} = event.result`.

src/api/notification/content-types/notification/schema.json
oriyomibadmus: There are plans in the future to sort all notifications into categories. Thought adding a tag field might be helpful

chungthuang: Maybe that's not needed if we do https://github.com/dev-launchers/strapiv4/pull/103/files#r1412896943.

src/api/comment/content-types/comment/lifecycles.js
chungthuang: I think the requirement from product is notification will only be sent to 1 user. In the case of comment, only the idea owner will be notified. In the case of applicant, only the owner of the opportunity will be notified. Too many notifications can cause fatigue and people will start ignoring them.

src/api/comment/content-types/comment/lifecycles.js
oriyomibadmus: @Enjoy2Live You are right! Done it!

src/api/idea-card/content-types/idea-card/lifecycles.js
oriyomibadmus: Yes, Done

src/api/notification/content-types/notification/schema.json
oriyomibadmus: @chungthuang Just made the changes to the notification collection schema to reflect these

chungthuang (approved):

src/api/comment/content-types/comment/lifecycles.js
Enjoy2Live: > I think the requirement from product is notification will only be sent to 1 user. In the case of comment, only the idea owner will be notified. In the case of applicant, only the owner of the opportunity will be notified. Too many notifications can cause fatigue and people will start ignoring them.

good point but my question is about the viability of changing the current solution we want to include multiple recipients of the same notification? (for example in the chat section someone might tag two people in the same comment)

Enjoy2Live: My guess is it's going to be an attribute in the notification collection but correct me if I'm wrong @oriyomibadmus .

btw happy new year! :smile:

Enjoy2Live (approved):
Awesome, that's a huge step.
Thank you guys

src/api/idea-card/content-types/idea-card/lifecycles.js
```diff
@@ -1,9 +1,30 @@
...
+        Content: "You created idea " + ideaName,
+        Collection: "IdeaCard",
+        ObjectID: id,
+        TimeCreated: timeCreated,
+        Read: false
+      },
+    });
+
+  }
+}
```
chungthuang, line 20: I think you have an extra `,` after the rebase.

src/api/idea-card/content-types/idea-card/lifecycles.js
```diff
@@ -1,9 +1,30 @@
...
+        Title:"A new idea has been created",
+        Content: "You created idea " + ideaName,
+        Collection: "IdeaCard",
+        ObjectID: id,
+        TimeCreated: timeCreated,
+        Read: false
+      },
+    });
+
+  }
```
chungthuang, line 19: To export multiple functions, you will need an `,` here.

PR #117, opened 2023-12-05
fix(chungthuang): change cookie to sameSite=none

chungthuang:


dbradham (approved):
Looks good according to the docs

PR #132, opened 2023-12-12
Make sensitive user data private fields

chrismferguson:
Please let me know if there's other fields in your opinion that should not private, I grabbed the ones that were the most urgent.


src/extensions/users-permissions/content-types/user/schema.json
```diff
@@ -102,7 +102,8 @@
       "type": "string"
```
chungthuang, line 102: I would say this is also private data.

chungthuang:
@chrismferguson are you familiar with postman? We have a test suite in https://github.com/dev-launchers/strapiv4/blob/main/integration-test/collection.json. We'll need to change some of the assertions in https://github.com/dev-launchers/strapiv4/blob/main/integration-test/collection.json#L110.

chrismferguson:
Oh, perfect, I've never used Postman before but I'm gonna try it today

chungthuang:
You'll also need to rebase on the main branch. The docker build was failing. https://github.com/dev-launchers/strapiv4/commit/f391141a36ad57c883675c72ec8ecf9f66f4f0a4 fixed it.

chrismferguson:
Still need to test with Postman, please wait.

PR #141, opened 2024-01-18
Node version upgrade 20

chrismferguson:


chungthuang (approved):

chungthuang:
Oh I didn't realize you were targeting the main branch. We should also change this on the main branch.

PR #143, opened 2024-01-20
fix(chrismferguson): Make sensitive user data private fields

chungthuang:


chungthuang:
Here's an example to make the commit history cleaner:
```
1. Checkout main branch
2. Create a new branch from main branch
3. Copy selected files from your original branch
$ git checkout privatized-user-fields integration-test/collection.json
Updated 1 path from 7ac99ab

$ git checkout privatized-user-fields src/extensions/users-permissions/content-types/user/schema.json
Updated 1 path from 7ac99ab

$ git checkout privatized-user-fields types/generated/contentTypes.d.ts
Updated 1 path from 7ac99ab
4. Commit the changes and create a PR
```
The result is https://github.com/dev-launchers/strapiv4/pull/143.

Issue #127, opened 2023-12-10
Remove TimeCapsule collection


chungthuang:

chrismferguson:
https://github.com/dev-launchers/strapiv4/pull/146

chrismferguson:
TimeCapsule collection was removed. Branch was merged into main without issue.

PR #146, opened 2024-02-09
fix(TimeCapsule): removed TimeCapsule collection 

chrismferguson:
This is the issue: https://github.com/dev-launchers/strapiv4/issues/127


chungthuang:
The changes looks good. Just one little thing, can we change the commit to follow this format https://github.com/semantic-release/semantic-release#commit-message-format? This is a small change so we can use fix release type.

chungthuang (approved):

becevka:
I noticed in other PRs in the repo the message is fix(authorname) but in the document it seems to be fix(domain-area) @chungthuang what approach we want to follow?

becevka (approved):
I think we need to rethink use of binary code in version control, like data.db. But I will table this discussion into Discord channel.

chungthuang:
> I noticed in other PRs in the repo the message is fix(authorname) but in the document it seems to be fix(domain-area) @chungthuang what approach we want to follow?

I usually do `fix(author name)`.

PR #152, opened 2024-02-25
build(npm): only install deps if engines are compatible

Enjoy2Live:
With this change npm will stubbornly refuse to install (or even consider installing)
any package﻿ unless the engine requirements specified in package.json are met.


becevka (approved):
Looks good to me. Thanks

PR #153, opened 2024-02-25
feat(dev-env): Improvements, Fixes, Documentation for local environment 

### LINKED ISSUE #149: Create documentation for authentication locally/dev mode
**Summary**
So far developers struggle to call authenticated routes in dev mode [`npm run develop`] as there's no documentation/guide page for that which makes them fallback to using staging environment of Strapi.

**Describe the solution you'd like**
We need a documentation page that describes the steps to create a user from frontend developer perspective and those steps end up with the developer successfully calling /api/users/me

## Acceptance Criteria
 - [ ] **Scenario: Create documentation for authentication locally/dev mode**
  - Given the developer is in dev mode using `npm run develop`
  - When the developer follows the new documentation on creating a user from the frontend perspective
  - Then the developer can successfully calls `/api/users/me` endpoint

### LINKED ISSUE #150: [Bug] "An error occurred while uploading the file." in local development mode
**Describe the bug**
when trying to upload assets to the media library in development environment you get
`An error occurred while uploading the file.`

**To Reproduce**
1. Go to content management
2. Click on profile [or any collection that has relationship to media manager]
3. Try to upload an image to the to that entity

Node.js returns this error in the terminal
```
strapiv4/node_modules/@azure/storage-blob/dist/index.js:23787
            throw new Error("Unable to extract containerName with provided information.");
                  ^

Error: Unable to extract containerName with provided information.
    at ContainerClient.getContainerNameFromUrl (/home/momaqbol/strapiv4/node_modules/@azure/storage-blob/dist/index.js:23787:19)
    at new ContainerClient (/home/momaqbol/strapiv4/node_modules/@azure/storage-blob/dist/index.js:22638:36)
    at BlobServiceClient.getContainerClient (/home/momaqbol/strapiv4/node_modules/@azure/storage-blob/dist/index.js:24363:16)
    at /home/momaqbol/strapiv4/node_modules/strapi-provider-upload-azure-storage/dist/index.js:38:47
    at Generator.next (<anonymous>)
    at /home/momaqbol/strapiv4/node_modules/strapi-provider-upload-azure-storage/dist/index.js:8:71
    at new Promise (<anonymous>)
    at __awaiter (/home/momaqbol/strapiv4/node_modules/strapi-provider-upload-azure-storage/dist/index.js:4:12)
    at handleUpload (/home/momaqbol/strapiv4/node_modules/strapi-provider-upload-azure-storage/dist/index.js:36:12)
    at Object.uploadStream (/home/momaqbol/strapiv4/node_modules/strapi-provider-upload-azure-storage/dist/index.js:91:24)

Node.js v18.4.0
```
**Expected behavior**
image gets uploaded

**Screenshots**
image

**Environment**
OS: PopOS
Browser: Librefox
Node: v18.4.0

becevka:
Closes #149 
Closes #150

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,9 +1,9 @@
  # Dev Launchers Strapi Service
 
 # Getting Started
-2. Run `npm run init` first time to install dependencies and initialize your local database
-3. Run `npm run develop`
-4. Go to http://localhost:1337/admin using email `local-admin@devlaunchers.org` and password `W&x5ZzOMtBCVt1YAUiuSzt~5`. 
+1. Run `npm run init` first time to install dependencies and initialize your local database
+2. Run `npm run develop`
+3. Go to http://localhost:1337/admin using email `local-admin@devlaunchers.org` and password `W&x5ZzOMtBCVt1YAUiuSzt~5`. 
 
 # Running from Docker
 Alternatively, you can run it with Docker. There are 2 make targets available to do this.
--- a/README.md
+++ b/README.md
@@ -1,10 +1,9 @@
  # Dev Launchers Strapi Service
 
 # Getting Started
-1. Copy the `.env.example` file into `.env`
-2. Run `npm install`
+2. Run `npm run init` first time to install dependencies and initialize your local database
 3. Run `npm run develop`
-4. Go to http://localhost:1337/admin to create an account
+4. Go to http://localhost:1337/admin using email `local-admin@devlaunchers.org` and password `W&x5ZzOMtBCVt1YAUiuSzt~5`. 
 
 # Running from Docker
 Alternatively, you can run it with Docker. There are 2 make targets available to do this.
@@ -32,5 +31,6 @@ Go to https://github.com/strapi/strapi/releases to find the latest version, then
 `package.json` to this version and run `npm install`.
 
 # Integration Test
-To run the integration test envrionment, set `NODE_ENV=test` in `.env` file. The login email is `integration-test@devlaunchers.org`
-password is `wQ46^BEsVbf9VD&D3ddSPCUP`. Data is stored in `integration-test/data.db`.
+To run the integration test envrionment, set `NODE_ENV=test` in `.env` file. 
+The login email is `integration-test@devlaunchers.org` and password is `wQ46^BEsVbf9VD&D3ddSPCUP`. 
+Data is stored in `integration-test/data.db`.
--- a/docs/local-setup.md
+++ b/docs/local-setup.md
@@ -26,6 +26,9 @@
 
 ## FrontEnd
 
+FrontEnd is available in our DevLaunchers Platform repository. 
+Please clone it using `git clone https://github.com/dev-launchers/dev-launchers-platform.git`.
+
 #### Start App
 1. Update `env.development` to point to local backend
 
@@ -42,11 +45,12 @@ NEXT_PUBLIC_GOOGLE_AUTH_URL=http://localhost:1337/api/connect/google
 
 4. Run the app's development server: `yarn workspace @devlaunchers/app dev`
 
-4. Go to http://localhost:3000 and sign up
+4. Go to http://localhost:3000 and sign up using your Google account
 
 5. After login, you should be able reproduce the steps from UI perspective
 
-# Cookies Issue
 
-After signing in DevLaunchers front end, login into Strapi Admin doesn't work due to some shared Cookies.
-It's suggested to open them in different browsers or use an incognito mode.
+# Additional Handy Commands
+
+`npm run init:db` -  removes project database and will setup all data again without reloading dependencies and no admin UI rebuild.
+`npm run init:permissions` - synchronizes only role permissions from init folder. This is useful for staging and production deployment to allow keep permissions in sync.
--- a/docs/local-setup.md
+++ b/docs/local-setup.md
@@ -1,0 +1,52 @@
+## Backend 
+
+#### Start App
+1. Run `npm run init` first time to install dependencies and initialize your local database
+3. Run `npm run develop`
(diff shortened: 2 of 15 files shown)


docs/local-setup.md
```diff
@@ -0,0 +1,52 @@
...
+
+7. Creates new Strapi admin account.
+
+8. Creates default interests and projects
+
+
+## FrontEnd
+
+#### Start App
+1. Update `env.development` to point to local backend
```
chungthuang, line 30: Should we suggest cloning the frontend repo?

init/config.js
```diff
@@ -0,0 +1,31 @@
+module.exports = {
+  google: {
+    enabled: true,
+    key: "27004093375-056i2bnbk9g1bpuhbe642asvcl9da5p5.apps.googleusercontent.com",
```
chungthuang, line 4: Where is this from? Let's not commit secrets into the repo.

init/setup.js
chungthuang, line 1: Is this idea borrowed from strapi? Can we link to the documentation?

init/permissions/public.js
```diff
@@ -0,0 +1,67 @@
+const initialize = {
+  find: { enabled: true, policy: "" },
+  create: { enabled: true, policy: "" },
+}
+const readOnly = {
+  find: { enabled: true, policy: "" },
+  findOne: { enabled: true, policy: "" },
+}
+
+module.exports = {
```
chungthuang, line 10: Should we do this to configure permissions in staging/production?

docs/local-setup.md
Enjoy2Live: @chungthuang don't think it's worth it because soon we're going to migrate to monorepo

README.md
```diff
@@ -1,10 +1,9 @@
  # Dev Launchers Strapi Service
 
 # Getting Started
-1. Copy the `.env.example` file into `.env`
-2. Run `npm install`
-3. Run `npm run develop`
-4. Go to http://localhost:1337/admin to create an account
+1. Run `npm run init` first time to install dependencies and initialize your local database
+2. Run `npm run develop`
+3. Go to http://localhost:1337/admin using email `local-admin@devlaunchers.org` and password `W&x5ZzOMtBCVt1YAUiuSzt~5`.
```
chungthuang, line 6: These steps also work for me!

init/config.js
Enjoy2Live: We have to do something about secrets, me and Wolfgang had a discussion about having this here and it seems like a scary idea but at the same time we have no other option as the auth flow depends on google.

We can delegate creating these secrets to the developer but this will create a bit of friction and I doubt front end devs would undergo the process of creating auth layer on their own google dashboard.

Enjoy2Live (approved):
@becevka Works like a charm, this will help alot since we're migrating to Vike.

Thanks a ton Wolfgang!

I approve this PR from frontend point of view but I know you guys have things to do for environment variables so no rushing

Enjoy2Live:
@alexanderyanthar check this out, it's our next step to try this with vike :smile:

docs/local-setup.md
```diff
@@ -0,0 +1,52 @@
...
+4. Run the app's development server: `yarn workspace @devlaunchers/app dev`
+
+4. Go to http://localhost:3000 and sign up
+
+5. After login, you should be able reproduce the steps from UI perspective
+
+# Cookies Issue
+
+After signing in DevLaunchers front end, login into Strapi Admin doesn't work due to some shared Cookies.
+It's suggested to open them in different browsers or use an incognito mode.
```
Enjoy2Live, line 52: is it possible to not have those cookies be shared? as far as I know that's not possible

init/config.js
chungthuang: Can we get this from an env var, and we give it to developers during onboarding?

becevka: Sorry, missed I need to submit my comment, and it was pending for two days. 
As I mentioned, ideally we want to have minimum friction during onboarding and the easiest approach is to use separate Google Auth Client App for local testing. Thus, I created new project in my google admin console to make sure we are not exposing prod/staging creds.

docs/local-setup.md
becevka: @Enjoy2Live 
Actually, it is possible to not have this issue if we agree to always refer to the FrontEnd as `http://127.0.0.1:3000` and to Backend as `http://127.0.0.2:1337`. That would be considered different domains. But we will need to update all references to `localhost`. Maybe that is something to discuss for monorepo integrations steps?

Enjoy2Live: So it's not possible to send cookies through a different port only, In this case only 3000? If that's the case then yes let's discuss it when migrating to monorepo

becevka: Correct. Cookies are domain-only specific. Same domain with multiple different ports will end up using all cookies for a domain. Another thought would be to see which cookie is that and maybe there's a setting to name it differently in Strapi. I will have a look.

docs/local-setup.md
becevka: Added

init/permissions/public.js
becevka: Added under `npm run init:permissions`

docs/local-setup.md
becevka: @Enjoy2Live @chungthuang 
Turns out it is not a cookie issue but the session token we set over from cookie to a header. Which is great for /api endpoints but not so well for /admin. https://github.com/dev-launchers/strapiv4/pull/153/files#diff-3075c669577ff744720d664128b19bf879a1fe1e48cc878639f0ac7f24253d5fR8
There's a dedicated extension for this specific issue, from which I borrowed the check to apply Authorization header only to /api endpoints https://github.com/bwyx/strapi-jwt-cookies/blob/main/middlewares/getter.js#L8

chungthuang: Great finding!

chungthuang (approved):

PR #155, opened 2024-02-25
feat(tests): Experiment with Jest Test Framework

becevka:
This is a draft proposal for test framework changes. 
I have also added Workflow action to demonstrate how the tests and coverage would work. Before final merging, I will need to update our test Workflow action to integrate with new test framework and migrate all existing and  integrations tests from postman to jest.

### DIFF
--- a/.github/workflows/jest.yaml
+++ b/.github/workflows/jest.yaml
@@ -13,7 +13,7 @@ jobs:
               uses: actions/setup-node@v3
               with:
                 node-version: '18.x'
-            - name: Run coverage report
+            - name: Run test and coverage reports
               uses: ArtiomTr/jest-coverage-report-action@v2
               env:
                 FRONTEND_URL: "not_used"
--- a/.github/workflows/jest.yaml
+++ b/.github/workflows/jest.yaml
@@ -7,7 +7,6 @@ on:
 jobs:
     test-report:
         runs-on: ubuntu-latest
-        continue-on-error: true
         steps:
             - uses: actions/checkout@v3
             - name: Use Node.js ${{ matrix.node-version }}
--- a/.github/workflows/jest.yaml
+++ b/.github/workflows/jest.yaml
@@ -10,11 +10,9 @@ jobs:
         steps:
             - uses: actions/checkout@v3
             - name: Use Node.js ${{ matrix.node-version }}
-              uses: actions/setup-node@v1
+              uses: actions/setup-node@v3
               with:
                 node-version: '18.x'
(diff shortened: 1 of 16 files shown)


.github/workflows/jest.yaml
```diff
@@ -0,0 +1,21 @@
+name: 'jest-test-report'
+on:
+  pull_request:
+    branches:
+      - becevka/jest-tests
```
becevka, line 5: Will remove this before merge.

package.json
```diff
@@ -7,7 +7,9 @@
     "develop": "strapi develop",
     "start": "strapi start",
     "build": "strapi build",
-    "strapi": "strapi"
+    "strapi": "strapi",
+    "test": "jest --runInBand",
```
chungthuang, line 11: Can we mention how to run the test sin README.

zar92:
So, are we not using postman anymore to test the APIs? Why are we changing from Postman to Jest?

chungthuang:
Hi @becevka, please meet @nikita-pardeshi-github who is the new lead for the QA team.

becevka:
@zar92 Ideally we will not continue with postman. The goal is to encourage writing tests and make it as easier and usable for developers as possible. 
Here are few points:

- we should avoid using prepared test.db, instead we can write code to create the data as needed inside the test suite
- postman collections are not readable if exported and it's a bit hard to understand which test belongs to which endpoint, thus code reviews of tests are very complicated
- jest is better integrated into VS Code and developers can run individual suites or tests from IDE
- jest has better assertion framework, which can be extend if needed
- jest is better integrated into github actions to show code coverage and failed test reports, for example check this report on my commit. It shows where test failed and which line wasn't covered https://github.com/dev-launchers/strapiv4/pull/155/commits/eb0a4b7cdefd0a555e68fdd8aab794f194ae3eaa Which makes much better for code reviews. And we can generate overall report in comments, like this https://github.com/dev-launchers/strapiv4/commit/eb0a4b7cdefd0a555e68fdd8aab794f194ae3eaa#comments

Here's an article providing more analysis, https://seccl.tech/blog/migrating-api-tests-from-postman-to-jest/ 
But generally speaking Jest is a very popular framework for running unit tests both in front end and backend JS world, and in combination with Supertest it works great for integration testing. 
And based on this draft experiment it can work easily with Strapi.

package.json
becevka: Added

becevka:
Closing in favor of https://github.com/dev-launchers/strapiv4/pull/163

PR #180, opened 2024-05-04
feat(comments): Backward compatibility and migration framework.

### LINKED ISSUE #158: [Bug] Frontend can misattribute authors of collections in strapi
**Describe the bug**
Currently the frontend can send requests to backend with "author" being a field in the request body and that field usually expects a number that refers to the id in the user collection.

This is a really bad way to handle ownership of content as anyone who knows a little bit about reverse engineering can maliciously use those endpoints and submit comments, ideas, etc while misattributing them to another user by altering the POST request body to include a different user.id than their's.

**To Reproduce**
1. Login in https://staging.devlaunchers.org
2. Navigate to https://staging.devlaunchers.org/ideaspace/workshop/254 or any other idea workshop
3. Submit a comment
4. Observe the request you made in network devtools
5. Mimic the request but with changing the payload to include a different "author" id
6. Successfully misattributed the comment to someone else

**Expected behavior**
"authorId" should point to the users-permissions.user id and should be inferred in strapiv4 based on the cookie "token" being supplied with the request and ONLY give authorship based on the currenly logged-in user token.

**Non-Functional Requirement**
Existing comments data should be migrated to fill the new "authorId" field based on current comment's author. 
Suggested to find a user by name and update authorId. After that, we can remove "author" field.

**Screenshots**
image

becevka:
Closes #158 Migration part

### DIFF
--- a/src/api/comment/content-types/comment/lifecycles.js
+++ b/src/api/comment/content-types/comment/lifecycles.js
@@ -3,7 +3,7 @@ module.exports = {
   beforeCreate(event) {
     // Set author to the user sending the request
     const ctx = strapi.requestContext.get();
-    event.params.data.authorId = ctx.state.user;
+    event.params.data.user = ctx.state.user;
     event.params.data.author = ctx.state.user.username;
   },
 
--- a/src/api/comment/content-types/comment/lifecycles.js
+++ b/src/api/comment/content-types/comment/lifecycles.js
@@ -4,7 +4,7 @@ module.exports = {
     // Set author to the user sending the request
     const ctx = strapi.requestContext.get();
     event.params.data.authorId = ctx.state.user;
-    console.log("create comment ctx", ctx)
+    event.params.data.author = ctx.state.user.username;
   },
 
   async afterCreate(event) {
--- a/src/api/comment/content-types/comment/schema.json
+++ b/src/api/comment/content-types/comment/schema.json
@@ -24,7 +24,7 @@
     "author": {
       "type": "string"
     },
-    "authorId": {
+    "user": {
       "type": "relation",
       "relation": "manyToOne",
       "target": "plugin::users-permissions.user",
--- a/src/migrators/comments.js
+++ b/src/migrators/comments.js
@@ -1,0 +1,27 @@
+module.exports = async (strapi) => {
+  strapi.log.info("Migrating comments");
+  const comments = await strapi.db.query("api::comment.comment").findMany({
+    where: {
+      authorId: null,
+    },
+    offset: 0,
+    limit: -1,
+  });
+  strapi.log.info(`Comments to be migrated: ${comments.length}`);
+  for (const comment of comments) {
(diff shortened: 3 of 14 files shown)


src/api/comment/content-types/comment/lifecycles.js
```diff
@@ -4,7 +4,7 @@ module.exports = {
     // Set author to the user sending the request
     const ctx = strapi.requestContext.get();
     event.params.data.authorId = ctx.state.user;
```
chungthuang, line 6: If `ctx.state.user` is the authorId, I would think the type is an integer. But `event.params.data.author = ctx.state.user.username` suggests its an object. Maybe this should be `ctx.state.user.Id` or some other field?

src/migrators/index.js
```diff
@@ -0,0 +1,8 @@
+module.exports = async (strapi) => {
+  const config = strapi.config.get("migrations");
+  if (config?.enabled === true && config?.migrators?.length > 0) {
+    for (const migrator of config.migrators) {
+      await require(`./${migrator}`)(strapi);
```
chungthuang, line 5: TIL you can do this!

src/migrators/comments.js
chungthuang, line 1: Do we plan to delete this file after it's run? Or we can add a version number like what DB migration scripts usually has and skip versions that has been applied.

src/migrators/index.js
Enjoy2Live: that's indeed sophisticated, I've seen it before but didn't know you can do dynamic value within the parenthesis.

src/api/comment/content-types/comment/lifecycles.js
becevka: Yeah, I originally thought similarly, to use just user.id, but for reference strapi needs the whole object. Since `author` already taken and we need to keep it as a string for backward compatibility maybe I should have a different field name to avoid confusion. `commenter` or `user` ? 🤔

src/migrators/comments.js
becevka: No need to delete, we can turn off migration or remove it from the list of migrators or control the list by env variable. https://github.com/dev-launchers/strapiv4/pull/180/files#diff-ff133665ecaacb3af753ccc05053349782668ced9af403ba7d39785615e57822R3

src/api/comment/content-types/comment/lifecycles.js
becevka: Ended up using `user`

src/api/comment/content-types/comment/schema.json
```diff
@@ -24,7 +24,7 @@
     "author": {
       "type": "string"
     },
-    "authorId": {
```
chungthuang, line 27: Quick search in the frontend repo shows that it's not using this field, so the renaming is safe.

chungthuang (approved):

dbradham (approved):

Issue #176, opened 2024-04-28
[Platform] [User Notifications][TUS]Subscriptions are created when user performs certain actions


becevka:
**Summary**
In order to support User Notifications, we need to subscribe users by default to the Idea of their interest.

**Describe the solution you'd like**
Subscription Collection should consist of the following fields:
- entityId, numeric id of the business-level entity (IdeaCard)
- entityType, enum value of business-level entity type (IdeaCard only for now)
- user, reference to the user subscribed to Business Entity
- createdDateTime, date when event was created
- active, boolean value, true by default, false when user unsubscribed
- unsubscribeDateTime, date when user deactivated subscription

When user creates idea, subscription for the new idea is automatically created.
When user posts comment to the idea, subscription for that idea is automatically created.
For now, we can disable API endpoints to Subscription.

## Acceptance Criteria
> Generated by [Zenhub AI](https://www.zenhub.com/ai)

- [ ] **Scenario: Subscription is created when user creates an idea**
  - Given a user creates an idea
  - When the idea is created
  - Then a subscription is automatically created for the user with the entityId of the idea, entityType as 'IdeaCard', and user as the user who created the idea
- [ ] **Scenario: Subscription is created when user posts a comment to an idea**
  - Given a user posts a comment on an idea
  - When the comment is posted
  - Then a subscription is automatically created for the user with the entityId of the idea, entityType as 'IdeaCard', and user as the user who posted the comment

becevka:
@tiagopazhs Please start with creating a Subscription collection.

tiagopazhs:
> @tiagopazhs Please start with creating a Subscription collection.

All rigth!

So, how it works? 

If I create the collection in my strapi admin. Then push to my branch.

If you pull this, you will see the collection in your strapi admin? Is it imediatly or do you need to run a migration?

tiagopazhs:
Why we have a created time instead of subscribed time?

tiagopazhs:
I didn't understand very weel about this two ones:

entityId, numeric id of the business-level entity (IdeaCard)
entityType, enum value of business-level entity type (IdeaCard only for now)

tiagopazhs:
I just created the PR here with the other fields: https://github.com/dev-launchers/strapiv4/pull/181

tiagopazhs:
So I have another question:
If I add a field to Subscription. What will happen with the other records in the database that doesn't have this new field.

becevka:
[...]

PR #181, opened 2024-05-11
feat(notifications): Create subscription collection

tiagopazhs:
## Required points discussed in the meeting of May-12:
- [x] Finalize the collection.
- [x] Update the collection names as similar as notification.
- [x] Add the behavior to create a subscription when user creates an idea.
- [x] Add test for idea creation and subscription create.
- [x] ~Fix this test ``expect(subscription.attributes.user.id).toBe(user.id)``~ (TBD in during Notification feature integration)
- [x] ~Verify that subscription are not available in API.~ (TBD in during Notification feature integration)

### DIFF
--- a/src/api/subscription/content-types/subscription/schema.json
+++ b/src/api/subscription/content-types/subscription/schema.json
@@ -8,7 +8,7 @@
     "description": ""
   },
   "options": {
-    "draftAndPublish": true
+    "draftAndPublish": false
   },
   "pluginOptions": {},
   "attributes": {
--- a/src/api/subscription/content-types/subscription/schema.json
+++ b/src/api/subscription/content-types/subscription/schema.json
@@ -35,6 +35,11 @@
     },
     "TimeCreated": {
       "type": "datetime"
+    },
+    "entityId": {
+      "type": "integer",
+      "required": true,
+      "unique": false
     }
   }
 }
--- a/src/api/subscription/content-types/subscription/schema.json
+++ b/src/api/subscription/content-types/subscription/schema.json
@@ -4,7 +4,8 @@
   "info": {
     "singularName": "subscription",
     "pluralName": "subscriptions",
-    "displayName": "Subscription"
+    "displayName": "Subscription",
+    "description": ""
   },
   "options": {
     "draftAndPublish": true
@@ -14,7 +15,7 @@
     "createdAt": {
       "type": "datetime"
     },
-    "unsubscribeAt": {
+    "TimeUnsubscribed": {
       "type": "datetime"
     },
     "active": {
@@ -25,6 +26,15 @@
       "relation": "manyToOne",
       "target": "plugin::users-permissions.user",
       "inversedBy": "subscriptions"
+    },
+    "entityType": {
+      "type": "enumeration",
+      "enum": [
+        "IdeaCard"
+      ]
+    },
+    "TimeCreated": {
+      "type": "datetime"
     }
   }
 }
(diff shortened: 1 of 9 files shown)


becevka:
Please add changes from types/generated/contentTypes.d.ts

src/api/subscription/content-types/subscription/schema.json
```diff
@@ -0,0 +1,30 @@
...
+  "info": {
+    "singularName": "subscription",
+    "pluralName": "subscriptions",
+    "displayName": "Subscription"
+  },
+  "options": {
+    "draftAndPublish": true
+  },
+  "pluginOptions": {},
+  "attributes": {
```
chungthuang, line 13: Do we want to allow user to select which entity type to subscribe https://github.com/dev-launchers/strapiv4/blob/main/src/api/notification/content-types/notification/schema.json#L27?

tiagopazhs:
This PR belongs to https://github.com/dev-launchers/strapiv4/issues/176

types/generated/contentTypes.d.ts
```diff
@@ -666,6 +666,11 @@ export interface PluginUsersPermissionsUser extends Schema.CollectionType {
       'oneToMany',
       'api::comment.comment'
     >;
+    subscriptions: Attribute.Relation<
+      'plugin::users-permissions.user',
+      'oneToMany',
+      'api::subscription.subscription'
+    >;
     createdAt: Attribute.DateTime;
```
tiagopazhs, line 674: @becevka is this file that you were talking about?
https://github.com/dev-launchers/strapiv4/pull/181#issuecomment-2106120482

tiagopazhs:
> Please add changes from types/generated/contentTypes.d.ts

Done!
You mean, this https://github.com/dev-launchers/strapiv4/pull/181#discussion_r1615252951 ?

What this file is used for?

src/api/subscription/content-types/subscription/schema.json
tiagopazhs: Yeah!

types/generated/contentTypes.d.ts
chungthuang: I think this is autogenerated content?

tiagopazhs:
So, I am trying to create an idea-card but I am getting this error:

image

My auth:
image

My idea-card permission:
image

Providers:
image

tiagopazhs:
## Required test that we must have to pass when we are running the tests:
Necessary to deal with this situation: the user is present in the post request but we cannot get it to check if it is correct:

`expect(subscription.attributes.user.id).toBe(user.id)`

image

chungthuang:
> ## Required test that we must have to pass when we are running the tests:
> 
> Necessary to deal with this situation: the user is present in the post request but we cannot get it to check if it is correct:
> 
> `expect(subscription.attributes.user.id).toBe(user.id)`
> 
> image

My guess is user is another collection, so you need to pass a query string `?populate=user` to get user data back. https://docs.strapi.io/dev-docs/api/rest/populate-select#population explains why.

tiagopazhs:
> * Fix this test `expect(subscription.attributes.user.id).toBe(user.id)`

Hey @becevka this is something that was already solved in another branch ?

becevka:
> > * Fix this test `expect(subscription.attributes.user.id).toBe(user.id)`
> 
> Hey @becevka this is something that was already solved in another branch ?

Not yet, that would be something separate, maybe next week

becevka (approved):
Looks good to me

Issue #173, opened 2024-04-28
[Platform] [User Notifications][TUS] Notifications are available and read-only within logged-in user context


becevka:
**Summary**
In order to support User Notifications, we need to assign them to the user and prevent unauthorized access to notification data.

**Describe the solution you'd like**
Notifications Collection should consist of a reference to a user, event and details on whether notification is still active, when it was created and when it was marked as read by user.
Notifications endpoint should allow only read notifications for the currently logged in user.
Notifications endpoint should not allow any modifications to notifications.
Notifications endpoint should allow to filter notifications by createDate and active fields.

## Acceptance Criteria
> Generated by [Zenhub AI](https://www.zenhub.com/ai)

 - [ ] **Scenario: Notifications are assigned to the user**
  - Given a user is logged in
  - When a notification is created
  - Then the notification should have a reference to the user
- [ ] **Scenario: Notifications endpoint only allows read access for the logged-in user**
  - Given a user is logged in
  - When the user requests their notifications
  - Then the user should only be able to read their notifications
- [ ] **Scenario: Notifications endpoint does not allow modifications to notifications**
  - Given a user is logged in
  - When the user attempts to modify their notifications
  - Then the user should receive an error or be unable to make changes
- [ ] **Scenario: Notifications endpoint allows filtering by createDate and active fields**
  - Given a user is logged in
  - When the user filters their notifications by createDate and active fields
  - Then the user should receive a filtered list of notifications based on the specified criteria

becevka:
@oriyomibadmus 
- Please add a user to existing Notification to experiment with strapi Policies.

PR #182, opened 2024-05-12
feat(notifications): Updated Notification Schema and Introduce Notification Policy to Filter Notification by User

oriyomibadmus:


src/api/notification/content-types/notification/schema.json
chungthuang, line 1: Why are we changing the schema?

src/api/notification/policies/isAuthenticated.js
```diff
@@ -0,0 +1,25 @@
...
+  // Check if user is authenticated
+  if (!ctx.state.user) {
+    console.log(ctx)
+    console.log("Not LOGGED IN")
+    return ctx.unauthorized('You must be logged in to access this resource.');
+  }
+
+  // Check if user is authorized to read notifications
+  const userId = ctx.state.user.id; // Get the ID of the logged-in user
+  const notificationUserId = ctx.request.body.user; // Get the user ID from the request body
```
chungthuang, line 13: Is user an object or just an integer?

src/api/notification/content-types/notification/schema.json
oriyomibadmus: In this updated logic, we are transforming the previous notification collection into a robust notification system. So we introduced 2 new collections subscribed users and events. Task #173 has more details.
The Notification collection would eventually have the following fields:
1. Event
2. User Reference (User subscribed for a specific notification)
3. CreatedDateTime
4. ReadDateTime

src/api/notification/policies/isAuthenticated.js
oriyomibadmus: Was testing the policy attachment to the api. This would also be edited

src/api/notification/content-types/notification/schema.json
chungthuang: > we are transforming the previous notification collection into a robust notification system
The description of this PR was to attach policy, so this kind of refactoring will usually be in a separate PR.

src/api/notification/policies/is-notification-owner.js
```diff
@@ -0,0 +1,15 @@
+
+module.exports = async (ctx) => {
+
+  const user = ctx.state.user;
+
+  ctx.request.query = ctx.request.query || {};
+  const query = ctx.request.query
+  query.filters = {
+    ...(query.filters || {}),
+    user: user.id
```
chungthuang, line 10: Is the goal of this function to set the query to the current logged in user, or check if the query is trying to access another user? The name `is-notification-owner` sounds like the later.

tests/notification.spec.ts
```diff
@@ -1,40 +1,60 @@
...
-            involveLevel: "minimum",
-            status: "workshopping",
+let user1Jwt, user2Jwt, notificationId;
+
+test.describe('/api/notification policy', () => {
+    test.beforeAll(async ({ request }) => {
+        const apiContext = api(request);
+
+        // Authenticate user1
+        const user1LoginResponse = await apiContext.post('/api/auth/local', {
```
becevka, line 12: So, instead of logging in two users we can check it for the same user. 
- we can create two notifications with two different users ids.
- then when we read notifications by currently logged in user, user should see only one notification
- to create notification we should use Strapi service, not POST /api/notifications. POST to /api/notification is disabled. You can grab Strapi object using `const strapi = await strapiConnect();` from https://github.com/dev-launchers/strapiv4/blob/main/tests/utils.ts#L60

tests/notification.spec.ts
```diff
@@ -1,40 +1,60 @@
...
+                Authorization: `Bearer ${user1Jwt}`,
+            },
         });
-        const notifications = await api(request).get("/api/notifications");
-        const notification = notifications.data
-        .find(item => (item.attributes.entityId === newComment.id && item.attributes.entityType === "Comment"));
-        expect(notification.attributes.Title).toBe("tester2 commented on Testing2");
-        expect(notification.attributes.Content).toBe("test comment");
+        const user1Notifications = await user1NotificationsResponse.json();
+
```
tiagopazhs, line 44: Is it necessary to convert the response to json?

If yes, Why the api is returning another response type different of json?

oriyomibadmus:
Since this PR is outdated, I will be closing it. The updated PR which contains both the creation of the notification policy and it's test case is here: https://github.com/dev-launchers/strapiv4/pull/194.

PR #184, opened 2024-05-25
feat(events): Support Events upon Idea Creation and Update

sky1122:
Issue： https://app.zenhub.com/workspaces/platform-team-631b73fa4dada357304fe853/issues/gh/dev-launchers/strapiv4/174

create event entity
create event when idea-card create and update


becevka:
Please name you PR to start with feat(events) or feat(notifications)

src/api/event/routes/event.js
```diff
@@ -0,0 +1,9 @@
+'use strict';
+
+/**
+ * event router
+ */
+
+const { createCoreRouter } = require('@strapi/strapi').factories;
```
becevka, line 7: In order to satisfy `Events endpoints should not be available through API.` should we disable routes or controller?

src/api/idea-card/content-types/idea-card/lifecycles.js
```diff
@@ -8,19 +8,62 @@ module.exports = {
...
-        Read: false
+        createdDateTime: timeCreated
       },
     });
-  }
+  },
+
+  async beforeUpdate(event) {
+    const { data, where, select, populate } = event.params;
+    const previousData = await strapi.entityService.findOne('api::idea-card.idea-card', where.id, { populate });
```
becevka, line 26: Do we need to pass through `populate` parameter?

src/api/idea-card/content-types/idea-card/lifecycles.js
```diff
@@ -8,19 +8,62 @@ module.exports = {
...
+    event.params.data.previousStatus = previousData.status;
+  },
+
+  async afterUpdate(event) {
+    const { result, params } = event;
+    const { previousStatus } = params.data;
+
+    if (result.status !== previousStatus) {
+      
+      const { id, ideaName, status, createdAt } = result;
```
becevka, line 38: Can we extract this block into a separate method to avoid duplication or simplify to create `title` and `content` conditionally and use them in one call?

becevka (changes requested):
https://github.com/dev-launchers/strapiv4/pull/184/files#file-tests-notification-spec-ts-L24 Please remake this test to use new events?

src/api/event/routes/event.js
chungthuang: I think that's a good idea. Will swagger still try to create documentation thinking the API is available?

src/api/idea-card/content-types/idea-card/lifecycles.js
chungthuang: And do we want to sanitize the input? Maybe we don't want to allow the caller to populate everything.

src/api/idea-card/content-types/idea-card/lifecycles.js
```diff
@@ -8,19 +8,62 @@ module.exports = {
...
+      const userName = params.data.userName || 'Unknown User';
```
chungthuang, line 39: Should this be `params.data.user.name`?

becevka: Probably, we need to use currently logged in user here. Please do create few tests to validate different scenarios.

src/api/event/routes/event.js
becevka: Let's keep API for now. We will need to make a decision  for both Events and Subscriptions. Probably the easiest approach would be to disable any permissions on API for all roles in deployment but enable readOnly for tests. But also we can drop API routes, though we will need to use internal Services in Strapi within tests.

oriyomibadmus (approved):

becevka (approved):
Since I requested changes, need to approve now :)

PR #186, opened 2024-06-08
feat(tests): Support for testcontainers and permissions

### LINKED ISSUE #91: [Bug] Invalid IDs cause the API to return 500 internal server error
**Describe the bug**
Sentry caught an exception that requests to https://apiv4-staging.devlaunchers.org/api/idea-cards/undefined triggered an exception
```
select "t0".*, "t0"."id", "t0"."created_by_id", "t0"."updated_by_id" from "public"."idea_cards" as "t0" where ("t0"."id" = $1) limit $2 - invalid input syntax for type integer: "undefined"
```

**To Reproduce**
Send a request to https://apiv4-staging.devlaunchers.org/api/idea-cards/undefined 

**Expected behavior**
The API returns 400 bad request.

becevka:
This PR does a few things, but they are intertwined a bit.

- Corrected  Documentation on required prerequisites of NodeJS and version manager
- Fixed Dockerfile.dev to align with documentation
- Added and documented support for running Postgres with testcontainers
- Removed outdated Postman tests
- Consolidated GitHib Actions
- Fixed #91 and Postgres error with comments migration
- Improved errors when collection is forbidden, instead of getting 500, we will have proper 403 error with details on which permissions were denied.


.github/workflows/test.yaml
```diff
@@ -1,33 +1,42 @@
 name: test
 on:
   pull_request:
-    workflow_dispatch:
+  workflow_dispatch:
```
chungthuang, line 4: Do we need to preserve the indentation to have valid yaml?

Dockerfile.dev
```diff
@@ -1,12 +1,12 @@
...
 WORKDIR /srv/app
 COPY ./package.json ./
 COPY ./package-lock.json ./
-RUN npm run init
 COPY config ./config
 COPY public ./public
+COPY init ./init
 COPY src ./src
-RUN npm run build
+RUN npm run init || true
```
chungthuang, line 10: Why don't we need the build step anymore?

.github/workflows/test.yaml
becevka: workflow_dispatch is one of the triggers, the previous indentation was incorrect. It should add 

Screenshot 2024-06-09 at 4 08 13 PM

Dockerfile.dev
becevka: Init command takes care of it https://github.com/dev-launchers/strapiv4/pull/186/files#diff-405535515db7762623f57c6a4877ff8bac7d28b8e899146ad82e6ca522362f4fR18

init/test-postgres.js
```diff
@@ -0,0 +1,23 @@
+const { PostgreSqlContainer } = require ('@testcontainers/postgresql');
+const { spawn } = require('node:child_process');
+
+const run = async () => {
+    const container = await new PostgreSqlContainer().start();
```
chungthuang, line 5: Seems like there is a way to specify port in java https://stackoverflow.com/questions/69132686/how-can-i-set-the-port-for-postgresql-when-using-testcontainers, maybe there's a way to do it in javascript, too. But this way is also fine.

src/api/idea-card/controllers/idea-card.js
```diff
@@ -1,9 +1,21 @@
...
  */
 
-const { createCoreController } = require('@strapi/strapi').factories;
+const { createCoreController } = require("@strapi/strapi").factories;
 
-module.exports = createCoreController('api::idea-card.idea-card');
+module.exports = createCoreController(
+  "api::idea-card.idea-card",
+  ({ strapi }) => ({
+    async findOne(ctx) {
```
chungthuang, line 14: Could this be a global middleware? This problem can happen to any endpoint.

tests/utils.ts
```diff
@@ -4,60 +4,55 @@ import Strapi, { LoadedStrapi } from "@strapi/strapi";
...
+            if (response.status() !== expectedStatus) {
                 console.log(`POST ${response.url()}`);
                 console.log(await response.status());
                 console.log(await response.text());
             }
-            expect(response.ok()).toBeTruthy();
+            expect(response.status()).toBe(expectedStatus);
             const data = await response.json();
-            return data.data;
+            return expectedStatus >= 400 ? data : data.data;
```
chungthuang, line 43: I didn't realize errors are not wrapped under `data`.

chungthuang:
Looks good to me overall.

chungthuang (approved):

init/test-postgres.js
becevka: It can be done by setting `.withExposedPorts({container: 5432, host: 5432})` but probably better to have random port to make sure on local machine there is no conflict with other application or postgres running on 5432.

src/api/idea-card/controllers/idea-card.js
becevka: Yep, but let do it in a separate PR

tests/utils.ts
becevka: Yep, response has `data, meta and error fields` https://docs.strapi.io/dev-docs/api/rest#requests

oriyomibadmus (approved):

PR #188, opened 2024-06-16
ops: Set revisionHistoryLimit to limit the number of pods in completed state

chungthuang:


becevka (approved):
Seems like default limit is 10. But I saw hundreds of them sometimes. Could that be something related to timeouts?

chungthuang:
> Seems like default limit is 10. But I saw hundreds of them sometimes. Could that be something related to timeouts?

I'm not sure yet. Maybe the pods status matters to. I saw a lot of them in `ContainerStatusUnknown`. I wonder if pods in that state are counted toward the revision limit.

Issue #179, opened 2024-04-28
[Bug] User can call /user/id API and access other user data


becevka:
**Describe the bug**

**To Reproduce**
1. Login to strapi using '/api/auth/local' and test user with id 1,  'tester'
2. Go to /api/users/2
3. User details are returned with response code 200

**Expected behavior**
Response code 403 or other error not allowing to get another user details

**ACCEPTANCE CRITERIA**
- PII is restricted to the currently authenticated user
- The IdeaSpace workshopping page no longer displays PII (current displays users' emails)
- The IdeaSpace comments no longer displays PII (currently displays users' full names)

becevka:
Good read https://stackoverflow.com/questions/74728231/why-all-users-in-strapi-have-access-to-update-all-users-profile

becevka:
- We will need to enable this as we should be able to access profiles of users who posted comment, or created notifications.
- The solution would be to add a middleware to select only configured fields in user and profile collection.

dbradham:
@chadcrotchett @chungthuang currently, user names and emails are displayed on the idea workshopping page. If this change is merged, IdeaSpace would need to update this page.

Chad, does it sound okay to you if we make a ticket in Platform board to use the display name instead of the emails or full legal names on the Idea Workshopping page? Or do you want this change to be made by the IdeaSpace team?

![Image](https://github.com/user-attachments/assets/4a35c102-4531-48ab-bca5-6fbbc21c6f89)


![Image](https://github.com/user-attachments/assets/3abd724c-92a9-4660-84ce-9f9ad536095e)

chadcrotchett:
I'd like to remove PII, such as email and full name....so, I think that's
what you are offering, right?

I'm on my phone now - but if you can send a screenshot of what it will look
like after the merge I could glance at it.

Thanks,

Chad

dbradham:
@chadcrotchett yes, we can remove PII from the Workshopping page.

- Instead of displaying the user's full name on the comments, we will update to show their display name
- Instead of displaying the user's email on the workshopping page, we will update to show their display name

chadcrotchett:
Sounds great!

PR #192, opened 2024-06-23
fix(test): Correct test run in github action

### LINKED ISSUE #190: [Bug] Reporting of failed tests no longer works
**Describe the bug**
After switch to testcontainers reporting of failed test no longer work through Github Actions

**To Reproduce**
1. Go to https://github.com/dev-launchers/strapiv4/pull/182/checks
2. See check passed
3. Open log and check Run Playwright tests section
4. See 1 test failed

**Expected behavior**
Check failed and Github notification is generated and report is shown on failing code

**Environment**
 Github

becevka:
Closes #190 
After some try and fail, I found that we can actually run Postgres Container, Strapi and tests within one context, which improves errors and code coverage reporting.


init/test-postgres.js
```diff
@@ -9,14 +9,14 @@ const run = async () => {
     process.env.TEST_CONTAINER= 'true';
 
     console.log('Running tests...');
-    const program = spawn('nyc', ['playwright', 'test']);
-
-    program.stdout.pipe(process.stdout);
-    program.stderr.pipe(process.stderr);
+    const program = spawn('nyc', ['playwright', 'test'], { stdio: 'inherit', env: process.env });
```
becevka, line 12: Pass env and inherit stdio to make sure github action would be able to read report from logs.

init/test-postgres.js
```diff
@@ -9,14 +9,14 @@ const run = async () => {
...
 
     program.on('close', (code) => {
-        console.log('Tests finished with code', code);  
-        container.stop().then(() => console.log('PG container stopped'));
+        console.log('Tests finished with exit code', code);  
+        container.stop().then(() => {
+            console.log('PG container stopped');
+            process.exit(code);
```
becevka, line 18: Passthrough exit code to make sure build fails.

tests/global-teardown.ts
```diff
@@ -2,6 +2,8 @@ import Strapi from "@strapi/strapi";
 import { cleanUpDatabase } from "../init/common";
 
 async function globalTeardown() {
+    // need to keep this line, otherwise github annotations will not work
```
becevka, line 5: This is very weird. Effectively Playwright reporter needs this variable, and I don't know how, but somehow printing this out here helps to process annotations report correctly. See https://github.com/microsoft/playwright/blob/main/packages/playwright/src/reporters/github.ts#L124
Compare logs in Github action without this line: https://github.com/dev-launchers/strapiv4/actions/runs/9637422370/job/26576618409?pr=192#step:6:96 
(error is not recognized as an error) and with this log
https://github.com/dev-launchers/strapiv4/actions/runs/9637479783/job/26576801796?pr=192#step:6:97
Maybe that teardown happens too fast or somethings, probably any logging would help, but I just found it by chance 🤷‍♂️

tests/global-setup.ts
```diff
@@ -7,7 +7,8 @@ async function globalSetup() {
     process.env.FRONTEND_URL = 'not_used';
 
     console.log('Bootstrapping test database...');
-    await bootstrapDatabase();
+    const strapiInstance = await bootstrapDatabase();
+    strapiInstance.listen();
```
becevka, line 11: Instead of starting strapi instance from playwright, we can just start listen here. Istanbul (code coverage tool) doesn't have access to code in the separately running server, but has access to code running in the main thread.

tests/idea-space.spec.ts
```diff
@@ -63,13 +65,13 @@ test.describe('comment migration', () => {
 
     test("should update comment author", async () => {
         const strapi = await strapiConnect();
-        const params = { populate: ["user"] };
-        await strapi.db.query("api::comment.comment").update({ where: { author: user.username }, data: { user: null } });
+        const params = { where: { id: commentId }, populate: ["user"] };
+        await strapi.db.query("api::comment.comment").update({ where: { id: commentId }, data: { user: null } });
         let comment = await strapi.db.query("api::comment.comment").findOne(params);
```
tiagopazhs, line 70: Hey @becevka ,
Why in this case the expected is null ?

package.json
```diff
@@ -11,9 +11,8 @@
     "start": "strapi start",
     "build": "strapi build",
     "strapi": "strapi",
-    "strapi:test": "strapi develop --no-build",
     "test": "playwright test",
-    "test:postgres": "node ./init/test-postgres.js",
+    "test:postgres": "TEST_CONTAINER=true nyc playwright test",
```
becevka, line 15: Instead of using script and spawn child process, we can start a container inside setup and stop it in teardown

tests/global-setup.ts
```diff
@@ -1,13 +1,23 @@
...
+        const container = await new PostgreSqlContainer().start();
+        console.log(`PG container started on port ${container.getPort()}`);
+        process.env.DATABASE_PORT = container.getPort().toString();
+        global.postgresContainer = container;
+    }
+
     console.log('Bootstrapping test database...');
-    await bootstrapDatabase();
+    const strapiInstance = await bootstrapDatabase();
+    await strapiInstance.listen();
```
becevka, line 19: Instead of starting strapi instance from playwright, we can just start listen here. Istanbul (code coverage tool) doesn't have access to code in the separately running server, but has access to code running in the main thread.

tests/global-teardown.ts
```diff
@@ -1,11 +1,17 @@
-import Strapi from "@strapi/strapi";
 import { cleanUpDatabase } from "../init/common";
 
 async function globalTeardown() {
-    if (process.env.TEST_CONTAINER !== 'true') {
-        await Strapi();
-        console.log('Deleting test database...');
-        await cleanUpDatabase(strapi);
+    if (process.env.TEST_CONTAINER === 'true') {
+        const container = global.postgresContainer;
```
becevka, line 5: Using `global` variables allows to share started Strapi and Container between setup and teardown routine.

tests/idea-space.spec.ts
```diff
@@ -63,8 +65,8 @@ test.describe('comment migration', () => {
 
     test("should update comment author", async () => {
         const strapi = await strapiConnect();
-        const params = { populate: ["user"] };
-        await strapi.db.query("api::comment.comment").update({ where: { author: user.username }, data: { user: null } });
+        const params = { where: { id: commentId }, populate: ["user"] };
```
becevka, line 68: Needed to add a filter for specific comment by id, otherwise test fails sometimes on Postgres, as comments are also created in https://github.com/dev-launchers/strapiv4/blob/main/tests/event.spec.ts#L52

tests/idea-space.spec.ts
becevka: on line 69 we updated comment and set user field to be Null. Here we verify user is null, run migration and then verify that user was found and comment was updated.

tiagopazhs:
Hey @becevka,

It is looking good now as we don't have errors in the following logs, rigth ?
image

I have two questions:

1) Why we are skipping 3 tests ?

2) Why don't we pull it into the `oriyomi/173-Notification-Available-LoggedIn-User` ( branch of `#182` that has the test failed ) before to pull it into `main` branch ?

tests/global-setup.ts
```diff
@@ -1,13 +1,23 @@
...
 
 async function globalSetup() {
     process.env.NODE_ENV = 'test';
     process.env.FRONTEND_URL = 'not_used';
 
+    if (process.env.TEST_CONTAINER === 'true') {
+        const container = await new PostgreSqlContainer().start();
+        console.log(`PG container started on port ${container.getPort()}`);
+        process.env.DATABASE_PORT = container.getPort().toString();
+        global.postgresContainer = container;
```
chungthuang, line 14: Where is `global` defined?

becevka:
@tiagopazhs 

1. We currently skip 2 tests as Notification changes have not yet been added to the main and 1 test while I am figuring out #179 
2. This change will not help oriyomi/173-Notification-Available-LoggedIn-User branch. I need to have changes approved separately and add it to main first. After it is approved and merged, we can rebase latest PRs with main. Pulling unapproved changes leads to a lot of potential issues.

tiagopazhs:
> 1. We currently skip 2 tests as Notification changes have not yet been added to the main and 1 test while I am figuring out [[Bug] User can call /user/id API and access other user data #179](https://github.com/dev-launchers/strapiv4/issues/179)

OK! Thanks for explain.

> 2. This change will not help oriyomi/173-Notification-Available-LoggedIn-User branch. I need to have changes approved separately and add it to main first. After it is approved and merged, we can rebase latest PRs with main. Pulling unapproved changes leads to a lot of potential issues.

Ok, I got it. 
This workflow make sense in this case?

1 - Finalize changes and aprove `becevka/190`
2 - Pull `main` branch into `becevka/190`
3 - Finalize changes and aprove `oriyomi/173`
4 - Pull `main` branch into `oriyomi/173`
5 - Pull  `becevka/190` branch into `oriyomi/173`
1 - Finalize changes and aprove `oriyomi/173`
2 - Pull `oriyomi/173` into `main`

Actually in this case, It looks not necessary but I mean in cases that we have a branchs that fix other branchs.

chungthuang (approved):

PR #194, opened 2024-07-01
feat(notifications): Notification With Events and Lifecycle

oriyomibadmus:


src/api/event/content-types/event/lifecycles.js
```diff
@@ -0,0 +1,37 @@
...
+
+        try {
+            // Fetch all users with active subscriptions for the given entityId and entityType
+            const subscriptions = await strapi.entityService.findMany('api::subscription.subscription', {
+                filters: {
+                    entityId: entityId,
+                    entityType: entityType,
+                    status: 'active',
+                },
+                populate: ['user'],
```
chungthuang, line 16: Do we need to populate with user data? I think without populate, strapi will still return the user ID which is what we need to generate notification.

src/api/notification/routes/custom-notification.js
```diff
@@ -0,0 +1,21 @@
+module.exports = {
+    routes: [
+      {
+        method: 'GET',
+        path: '/notifications',
+        handler: 'notification.find',
```
chungthuang, line 6: The frontend will use this endpoint to list notifications but filter by user ID?

src/api/notification/policies/is-notification-owner.js
chungthuang, line 1: Maybe we should call this function enforce-notification-owner because we are forcing the query to filter by the logged in user.

VictorDiniz89 (approved):

src/api/event/content-types/event/lifecycles.js
becevka: FYI. Strapi doesn't include any data on the relation without populate parameter. This is done to correctly control published and allowed data. And that is probably the reason why even one-to-one relationship like for example user to profile are actually stored in many-to-many intermediary table. 
https://docs.strapi.io/dev-docs/api/rest/populate-select#population 
```The REST API by default does not populate any type of fields, so it will not populate relations, media fields, components, or dynamic zones unless you pass a populate parameter to populate various field types.```
One would think that for EntityService, they could have make it more straightforward, but from what I saw it behaves the same way, without `populate` parameter in the example above subscription.user will be undefined.

src/api/notification/policies/is-notification-owner.js
oriyomibadmus: Yes, the function has been rename to "notification-owner-filter"

becevka (approved):
LGTM

chungthuang (approved):

PR #196, opened 2024-07-09
feat(permissions): Restrict user and profile details based on allowed fields

becevka:
Addresses #179 
Please ignore commits from already squashed branch, I needed those while was waiting on PR approval. 
Any suggestions how we can simplify this middleware are welcome.

### DIFF
--- a/src/middlewares/restrict-user-details.js
+++ b/src/middlewares/restrict-user-details.js
@@ -1,0 +1,114 @@
+/***
+ * Main algorithm here: take the response and introspect all attributes, if any attribute is a relation, we take that object and introspect its attributes. 
+ * When relation points to a user or profile, we replace that object with shrunk version, including only allowed fields.
+ * Using private fields in Strapi Schema is not an option, as they are also not available for the user calling `/api/users/me` or `/api/users/:selfId`.
+ * The goal of this middleware is to restrict those fields for non-owner users.
+ * 
+ * Metadata is not properly documented in Strapi. 
+ * It is a Mapa Map-alike object where key is a modelName and value is MetaModel, which contains almost all the same info you can find under `content-types/schema.json`. 
+ * See https://github.com/strapi/strapi/blob/develop/packages/core/database/src/types/index.ts#L216 and https://github.com/strapi/strapi/blob/develop/packages/core/database/src/metadata/metadata.ts
+ */
+
+module.exports = (config, { strapi }) => {
+  return async (ctx, next) => {
+    await next();
+    try {
+      // Only need to filter if there is a response body
+      if (ctx.request.method === "GET" && ctx.request.url.includes("/api/") && ctx.response.body) {
+        const handler = ctx.state.route.handler; // for example: "user.me" or "api::profile.profile.findOne"
+        const user = ctx.state.user;
+        const idx = handler.lastIndexOf(".");
+        const modelName = handler.substring(0, idx); // for example: "user" or "api::profile.profile"
+        const action = handler.substring(idx + 1); // for example: "me" or "findOne"
+        const metadata = strapi.db.metadata;
+
+        const fieldsRestricter = new FieldsRestricter(metadata, config);
+
+        ctx.response.body = fieldsRestricter.restrictFields(ctx.response.body, modelName, action, user);
+      }
+    } catch (error) {
+      console.error(error);
+    }
+  };
+};
+
+class FieldsRestricter {
+  constructor(metadata, config) {
+    this.metadata = metadata;
+    this.config = config;
+  }
+
+  findAndRestrictUserDetails(item, modelName) {
+    if (modelName === "user") {
+      return this.restrictNonWrappedItem(item, this.config.allowedUserFields);
+    }
+    if (modelName === "plugin::users-permissions.user") {
+      return this.restrictWrappedItem(item, this.config.allowedUserFields);
+    }
+    if (modelName === "api::profile.profile") {
+      return this.restrictWrappedItem(item, this.config.allowedProfileFields);
+    }
+    if (!this.metadata.has(modelName)) {
+      return item;
+    }
+    const model = this.metadata.get(modelName);
+    if (model && item?.data) {
+      if (Array.isArray(item.data)) {
+        item.data.forEach((element) => {
+          this.reviewModelAttributes(element.attributes, model);
+        });
+      } else {
+        this.reviewModelAttributes(item.data.attributes, model);
+      }
+    }
+    return item;
+  }
+
+  reviewModelAttributes(item, model) {
+    for (let attribute in model.attributes) {
+      const attributeData = model.attributes[attribute];
+      if (attributeData.type === "relation" && item[attribute]) {
+        item[attribute] = this.findAndRestrictUserDetails(item[attribute], attributeData.target);
+      }
+    }
+  }
+
+  restrictWrappedItem(item, allowedFields) {
+    const restrictedItem = { data: { id: item.data.id, attributes: {} } };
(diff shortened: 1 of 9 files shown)


src/middlewares/restrict-user-details.js
```diff
@@ -0,0 +1,102 @@
...
+    if (modelName === "user" && action === "me") {
+      return response;
+    }
+    if (modelName === "user" && action === 'findOne' && response?.id === user.id) {
+      return response;
+    }
+    if (modelName === "api::profile.profile" && action === 'findOne' && response?.data?.attributes?.user?.data?.id === user.id) {
+      return response;
+    }
+    return this.findAndRestrictUserDetails(response, modelName);
```
becevka, line 98: Main algorithm here: take the response and introspect all attributes, if any attribute is a relation, we take that object and introspect its attributes. When relation points to `user` or `profile`, we replace that object with shrunk version, including only allowed fields.

src/middlewares/restrict-user-details.js
```diff
@@ -0,0 +1,102 @@
...
+  }
+
+  restrictFields(response, modelName, action, user) {
+    if (modelName === "user" && action === "me") {
+      return response;
+    }
+    if (modelName === "user" && action === 'findOne' && response?.id === user.id) {
+      return response;
+    }
+    if (modelName === "api::profile.profile" && action === 'findOne' && response?.data?.attributes?.user?.data?.id === user.id) {
```
becevka, line 95: If profile was requested with additionally populated user, we can compare it with current user and return full profile, if current user is an owner of the profile. I noticed, that frontend doesn't call `profiles` API directly, maybe we can skip this check and always return limited version of the profile data?

src/middlewares/restrict-user-details.js
```diff
@@ -0,0 +1,102 @@
+module.exports = (config, { strapi }) => {
+  return async (ctx, next) => {
+    await next();
+    try {
+      if (ctx.request.method === "GET" && ctx.request.url.includes("/api/")) {
```
chungthuang, line 5: Are we checking `/api/` to filter out requests from admin dashboard?

src/middlewares/restrict-user-details.js
```diff
@@ -0,0 +1,102 @@
...
+  constructor(metadata, config) {
+    this.metadata = metadata;
+    this.config = config;
+  }
+
+  findAndRestrictUserDetails(item, modelName) {
+    if (modelName === "user") {
+      return this.restrictNonWrappedItem(item, this.config.allowedUserFields);
+    }
+    if (modelName === "plugin::users-permissions.user") {
```
chungthuang, line 33: When will we see this modelName instead of just `user`?

src/middlewares/restrict-user-details.js
chungthuang: If the profile API is not called directly, should we remove the router?

src/middlewares/restrict-user-details.js
```diff
@@ -0,0 +1,102 @@
...
+  return async (ctx, next) => {
+    await next();
+    try {
+      if (ctx.request.method === "GET" && ctx.request.url.includes("/api/")) {
+        const handler = ctx.state.route.handler;
+        const user = ctx.state.user;
+        const idx = handler.lastIndexOf(".");
+        const modelName = handler.substring(0, idx);
+        const action = handler.substring(idx + 1);
+        const metadata = strapi.db.metadata;
```
chungthuang, line 11: What are examples of metadata?

src/middlewares/restrict-user-details.js
```diff
@@ -0,0 +1,102 @@
+module.exports = (config, { strapi }) => {
+  return async (ctx, next) => {
+    await next();
+    try {
+      if (ctx.request.method === "GET" && ctx.request.url.includes("/api/")) {
+        const handler = ctx.state.route.handler;
+        const user = ctx.state.user;
+        const idx = handler.lastIndexOf(".");
+        const modelName = handler.substring(0, idx);
+        const action = handler.substring(idx + 1);
```
chungthuang, line 10: Maybe we can add an example as comment.

src/middlewares/restrict-user-details.js
```diff
@@ -0,0 +1,102 @@
+module.exports = (config, { strapi }) => {
```
chungthuang, line 1: Can you add a comment explaining why it's not enough to mark the fields private (under advanced https://docs.strapi.io/user-docs/content-type-builder/configuring-fields-content-type)?

src/middlewares/restrict-user-details.js
becevka: Probably, yes. I will add this to the list proposed permissions here https://docs.google.com/spreadsheets/d/1de3jpktjxFUdEUbmPmtizclECpFfak7Y_1GC3j-QfH0/edit?gid=1553006189#gid=1553006189

src/middlewares/restrict-user-details.js
becevka: Correct

src/middlewares/restrict-user-details.js
becevka: When we call `/api/users/1` we will get handler `user.findOne`, but let say we call `api/idea-cards/1?populate=ideaOwner`, that ideaOwner will be a relation type and will target not `user` but  `plugin::users-permissions.user`. Other modelNames are more consistent `api::profile.profile` is the same in both handler and relation target type.

src/middlewares/restrict-user-details.js
becevka: Metadata is a Map-alike object where key is a modelName and value is MetaModel, which contains almost all the same info you can find under `content-types/schema.json` See https://github.com/strapi/strapi/blob/develop/packages/core/database/src/types/index.ts#L216 and https://github.com/strapi/strapi/blob/develop/packages/core/database/src/metadata/metadata.ts

src/middlewares/restrict-user-details.js
becevka: Done

src/middlewares/restrict-user-details.js
becevka: Done

chungthuang (approved):

dbradham:
@nikita-pardeshi-github @chungthuang @tiagopazhs @oriyomibadmus is anyone here familiar with the file that has a conflict? I'd like to get this merged, as Yacine has fixed some breaking changes on the frontend side, so we can now merge and deploy this change.

chungthuang (approved):

src/middlewares/restrict-user-details.js
```diff
@@ -0,0 +1,114 @@
...
+ * It is a Mapa Map-alike object where key is a modelName and value is MetaModel, which contains almost all the same info you can find under `content-types/schema.json`. 
+ * See https://github.com/strapi/strapi/blob/develop/packages/core/database/src/types/index.ts#L216 and https://github.com/strapi/strapi/blob/develop/packages/core/database/src/metadata/metadata.ts
+ */
+
+module.exports = (config, { strapi }) => {
+  return async (ctx, next) => {
+    await next();
+    try {
+      // Only need to filter if there is a response body
+      if (ctx.request.method === "GET" && ctx.request.url.includes("/api/") && ctx.response.body) {
```
chungthuang, line 17: @becevka I added a check for non empty `ctx.response.body`, otherwise https://github.com/dev-launchers/strapiv4/pull/196/files#diff-020f47be16701da9ac1d16017015d4d6d586e314e61850aa215c4c801450563eR27 will set it to an empty body and change the status code from 404 to 204.

Issue #191, opened 2024-06-23
[FEAT] Project API Functionality Coverage


becevka:
**Summary**
To improve stability of Project page backend API we should improve code coverage of Project API code.

**Describe the solution you'd like**
Review current use cases and custom code within Project API code and create tests to support 

**Acceptance criteria**
> Generated by [Zenhub AI](https://www.zenhub.com/ai)

 - [ ] **Scenario: Code coverage improvement for Project API functions**
  - Given the current Project API codebase
  - When additional test cases are created to cover all possible use cases
  - Then the code coverage should be improved, ensuring that all functions and edge cases are tested
- [ ] **Scenario: Testing custom code within Project API**
  - Given custom code has been implemented in the Project API
  - When tests are written to specifically target these custom implementations
  - Then the custom code should be thoroughly tested, ensuring its functionality and stability

PR #197, opened 2024-07-13
feat: 191 project api functionality coverage not_to_merge

nikita-pardeshi-github:
_Feature_ : https://github.com/dev-launchers/strapiv4/issues/191

_Implementation details_

1. Created project in common.js file under init folder.
2. Added test in project.spec.ts to test explicit else case.

@becevka Could you please review it once?


init/common.js
```diff
@@ -5,6 +5,7 @@ const _ = require("lodash");
 const authenticatedRolePermissions = require("./permissions/authenticated");
 const publicRolePermissions = require("./permissions/public");
 const config = require("./config");
+const { default: init } = require("@strapi/strapi/dist/Strapi");
```
becevka, line 8: Please don't import from distribution folders. You are not using this import

init/common.js
```diff
@@ -145,7 +153,7 @@ const setupGoogleAuthProvider = async () => {
 
 const createInterests = async (userId) => {
   const service = strapi.service("api::interest.interest");
-  let existing = await service.findOne({});
+  let existing = await service.findOne(userId);
```
becevka, line 156: That's not how findOne works, if you want to find existing interest by userId, you will need to use findMany with filters parameter https://docs.strapi.io/dev-docs/api/entity-service/crud#findmany

init/common.js
```diff
@@ -189,6 +197,18 @@ const createProject = async (interestId, userId) => {
   });
 };
 
+const createProjectWithoutTeam = async (interestId, userId) => {
```
becevka, line 200: You could modify previous method `createProject` and pass project object and boolean flag `addTeam`

init/config.js
```diff
@@ -47,4 +55,16 @@ module.exports = {
     signupFormUrl: "http://localhost:3000",
     isListed: true,
   },
+  projectWithoutTeam: {
+    slug: "test-dev-recruiting",
```
becevka, line 59: Let's use some different names. For example, you can use 'dev-launchers-platform' project from https://apiv4.devlaunchers.org/api/projects

tests/project.spec.ts
```diff
@@ -1,5 +1,5 @@
-import { test, expect } from '@playwright/test';
-import { api } from './utils';
+import { test, expect, request } from '@playwright/test';
```
becevka, line 1: No need to import request directly

becevka (commented):
Left few comments to improve overall code readability. Also, please update PR name to follow scheme `feat(tests): description`.

init/common.js
```diff
@@ -189,6 +197,18 @@ const createProject = async (interestId, userId) => {
...
+  const service = strapi.service("api::project.project");
+  let existing = await service.findOne(interestId);
```
chungthuang, line 202: This looks a bit weird that we are trying to find project by interest ID?

nikita-pardeshi-github: Thank you pointing out. Updated.

init/common.js
```diff
@@ -162,28 +167,31 @@ const createInterests = async (userId) => {
...
-    leaders: [
-      {
-        leader: userId,
-        role: "Project Manager",
-      },
-    ],
-  };
-  existing = await service.create({ data: config.project });
+  project.interests = [interestId];
+  if(!addTeam){
```
becevka, line 177: Probably need to reverse logic here

init/config.js
```diff
@@ -47,4 +47,16 @@ module.exports = {
     signupFormUrl: "http://localhost:3000",
     isListed: true,
   },
+  projectWithoutTeam: {
+    slug: "dev-launchers-platform",
+    title: "Dev Recruit",
```
becevka, line 52: Please use all other fields from https://apiv4.devlaunchers.org/api/projects

init/common.js
chungthuang: @nikita-pardeshi-github should we remove this then?

init/common.js
```diff
@@ -162,30 +167,40 @@ const createInterests = async (userId) => {
   return interestId;
 };
 
-const createProject = async (interestId, userId) => {
-  const service = strapi.service("api::project.project");
-  let existing = await service.findOne({});
+const createProject = async (interestId, userId, project, addTeam=false) => {
+  let existing = await strapi.entityService.findMany("api::project.project", {filters: {slug: project.slug}});
+  existing = existing[0];
```
chungthuang, line 172: Should we use the findOne method if we are only using the first?

tests/user.spec.ts
```diff
@@ -9,7 +9,7 @@ test.describe('/api/users', () => {
         expect(user.username).toBe(config.user.username);
         expect(user.email).toBe(config.user.email);
         expect(user.confirmed).toBe(config.user.confirmed);
-        expect(user.projects.length).toBe(1);
+        expect(user.projects.length).not.toBe(0);
         const project = user.projects[0];
```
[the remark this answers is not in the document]
tiagopazhs: Also another question.
In this scenario is it possible the user have no projects ?
Because if it is possible and we check the length of projects (and projects is nulll) it could return an error.

chungthuang:
@nikita-pardeshi-github https://github.com/dev-launchers/strapiv4/commit/53a9d822895f8b27c9bc56bfb65d9525ef80046d should fix the flaky test. Can you rebase?

nikita-pardeshi-github:
> @nikita-pardeshi-github [53a9d82](https://github.com/dev-launchers/strapiv4/commit/53a9d822895f8b27c9bc56bfb65d9525ef80046d) should fix the flaky test. Can you rebase?

Yes I will. Thank you @chungthuang

tests/user.spec.ts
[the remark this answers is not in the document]
nikita-pardeshi-github: Hi @tiagopazhs Thank you pointing out. Yes you are right, in real world it may happen. But in our test env this will not be the case anytime. If you see in common.js file we are creating project as part of setup and projects are assigned to the config.user. Also in user.spec.ts we are using the same user from config.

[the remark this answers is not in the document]
nikita-pardeshi-github: Yes, Also user can have more than one project. That case we have covered as part of test setup process.

init/common.js
chungthuang: @nikita-pardeshi-github I think Wolfgang means this should be `if(addTeam)`.

PR #195, opened 2024-07-07
#124 fix remove point collection

tiagopazhs:
## Changes
- First of all I deleleted refer with users.
  - I checked and the unique relation of point was really with users.
- Then I deleted Point Collection from UI.
- Check if point were deleted from databases tables:
  - Before:
    - It was empty but it existed:
    - image
  - Current:
    - Now it doesn't exists:
    - image

## Important
- I saw that we refer point in the current path: 
`/init/permissions/authenticated.js:67`
Should we remove that ?

- Why is `types/generated/contentTypes.d.ts` generating `authorId` ?

PR #206, opened 2024-08-09
Removing point collection

tiagopazhs:
Trying to fix branch conflicts after we start using user insead of author id.

Plase considerar the conversation on: from https://github.com/dev-launchers/strapiv4/pull/195


src/api/comment/content-types/comment/schema.json
```diff
@@ -29,6 +29,12 @@
       "relation": "manyToOne",
       "target": "plugin::users-permissions.user",
       "inversedBy": "comments"
+    },
```
chungthuang, line 32: I think something went wrong during the rebase, because authorId is removed from the main branch https://github.com/dev-launchers/strapiv4/blob/main/src/api/comment/content-types/comment/schema.json

tiagopazhs: Yeah, I don't know very well what is wrong.

So to rebase main, is this the rigth command ?

```
➜  strapiv4-api git:(tiago/124-fix-remove-point-collection) git rebase main

Current branch tiago/124-fix-remove-point-collection is up to date.
```

chungthuang: The command looks right. One possibility is that your local main branch was not updated. You can also fix this by removing the authorId field manually.

chungthuang (approved):

PR #209, opened 2024-08-12
Added role category and role type field for oppotunities

ajaywayase:


src/api/opportunity/content-types/opportunity/schema.json
```diff
@@ -55,6 +55,33 @@
...
+        "Information Architect",
+        "Lead Developer",
+        "Back-End Developer",
+        "Front-End Developer",
+        "QA Lead",
+        "QA Tester",
+        "Volunteer Coordinator",
+        "Social Media Manager"
+      ],
+      "required": true
```
chungthuang, line 84: What would happen to existing data that doesn't have this field? Should we add a default and then we can fix it manually. https://docs.strapi.io/user-docs/content-type-builder/configuring-fields-content-type#-enumeration says we can have a default under `Advanced settings`.

ajaywayase: Changes Done as suggested

chungthuang (approved):

PR #211, opened 2024-08-14
Opportunity api test coverage

nikita-pardeshi-github:
@chungthuang Creating a draft pr just to know whatever I have updated till now is correct. I would like to know if this test case suffice for opportunity api. Could you please take a look at it once?

### DIFF
--- a/init/common.js
+++ b/init/common.js
@@ -61,10 +61,8 @@ const bootstrapDatabase = async () => {
   // creating project without associating team
   await createProject(interestId, userId, config.projectWithoutTeam, false);
   console.log("Project created without associated team");
-  await createOpportunity(interestId, config.firstOpportunity);
-  console.log("First opportunity created");
-  await createOpportunity(interestId, config.secondOpportunity);
-  console.log("Second opportunity created");
+  await createOpportunities(interestId);
+  console.log("Opportunities created");
   return instance;
 };
 
@@ -206,17 +204,27 @@ const createProject = async (interestId, userId, project, addTeam=true) => {
   });
 };
 
-const createOpportunity = async (interestId, opportunity) => {
-    let existing = await strapi.entityService.findMany("api::opportunity.opportunity", {filters: {title: opportunity.title}});
-    existing = existing[0];
+const createOpportunities = async (interestId) => {
+
+  for (let index = 0; index < config.opportunity.length; index++) {
+    try{
+      let existing = await strapi.entityService.findMany("api::opportunity.opportunity", {filters: {title: config.opportunity[index].title}});
+      existing = existing[0];
+    }
+    catch (e){
+      console.log(`Could not create opportunity due to an error ${e}`)
+      continue
+    }
+    
     if (existing) {
       return;
     }
-  
-    opportunity.interests = [interestId];
+    
+    config.opportunity[index].interests = [interestId];
     
     const service = strapi.service("api::opportunity.opportunity");
-    existing = await service.create({ data: opportunity });
+    await service.create({ data: config.opportunity[index] });
+  }
 };
 
 module.exports = {
@@ -232,5 +240,5 @@ module.exports = {
   setupGoogleAuthProvider,
   createInterests,
   createProject,
-  createOpportunity,
+  createOpportunities,
 };
--- a/init/common.js
+++ b/init/common.js
@@ -61,8 +61,8 @@ const bootstrapDatabase = async () => {
   // creating project without associating team
   await createProject(interestId, userId, config.projectWithoutTeam, false);
   console.log("Project created without associated team");
-  await createOpportunity(interestId);
-  console.log("Opportunity created");
+  await createOpportunities(interestId);
+  console.log("Opportunities created");
   return instance;
 };
 
@@ -204,17 +204,21 @@ const createProject = async (interestId, userId, project, addTeam=true) => {
   });
 };
 
-const createOpportunity = async (interestId) => {
-  let existing = await strapi.entityService.findMany("api::opportunity.opportunity", {filters: {title: config.opportunity.title}});
-  existing = existing[0];
-  if (existing) {
-    return;
-  }
-
-  config.opportunity.interests = [interestId];
+const createOpportunities = async (interestId) => {
+  let opportunity = config.opportunity
+  for (let index = 0; index < config.opportunityTitles.length; index++) {
+    opportunity.title = config.opportunityTitles[index]
+    let existing = await strapi.entityService.findMany("api::opportunity.opportunity", {filters: {title: opportunity.title}});
+    existing = existing[0];
+    if (existing) {
+      return;
+    }
   
-  const service = strapi.service("api::opportunity.opportunity");
-  existing = await service.create({ data: config.opportunity });
+    config.opportunity.interests = [interestId];
+    
+    const service = strapi.service("api::opportunity.opportunity");
+    existing = await service.create({ data: config.opportunity });
+  }
 };
 
 module.exports = {
@@ -230,5 +234,5 @@ module.exports = {
   setupGoogleAuthProvider,
   createInterests,
   createProject,
-  createOpportunity,
+  createOpportunities,
 };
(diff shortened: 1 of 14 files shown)


init/common.js
```diff
@@ -202,6 +204,19 @@ const createProject = async (interestId, userId, project, addTeam=true) => {
   });
 };
 
+const createOpportunity = async (interestId) => {
+  let existing = await strapi.entityService.findMany("api::opportunity.opportunity", {filters: {title: config.opportunity.title}});
+  existing = existing[0];
```
chungthuang, line 209: Since we only need the first opportunity, we can use `strapi.entityService.findOne`.

tests/opportunity.spec.ts
```diff
@@ -0,0 +1,23 @@
+import { test, expect } from '@playwright/test';
+import { api } from './utils';
+import * as config from '../init/config';
+
+test.describe('/api/opportunities', () => {
```
chungthuang, line 5: We should also test the get one endpoint (`/api/opportunities/:id`).

chungthuang:
> @chungthuang Creating a draft pr just to know whatever I have updated till now is correct. I would like to know if this test case suffice for opportunity api. Could you please take a look at it once?

This looks like a good start! We should also test the other endpoints called by the frontend https://github.com/dev-launchers/dev-launchers-platform/blob/master/packages/utility/agent.ts#L118-L136.
- get: calls the list endpoint but filter by project
- getById: calls the get one endpoint

init/common.js
nikita-pardeshi-github: @chungthuang  Thank you for your review. findOne with filteres not working. https://docs.strapi.io/dev-docs/api/entity-service/filter. Is there any alternate way to filter by title?

nikita-pardeshi-github:
@chungthuang added test case to fetch opportunity filtered by id. Could you please review once?

init/config.js
```diff
@@ -61,4 +61,19 @@ module.exports = {
...
+      {
+        "id": 0,
+        "expectation": "test expectation"
+      }
+    ],
+    isHidden: true,
+    roleCategory: "Development",
+    roleType: "Back-End Developer"
+  },
+  opportunityTitles : ["test opportunity", "test second opportunity"]
```
chungthuang, line 78: This looks a bit unnatural. Can we create 2 opportunities instead of the same opportunity, but 2 titles.

tests/opportunity.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+
+    test("Get all opportunities", async ({ request }) => {
+        const opportunityData = await api(request).getData("/api/opportunities");
+
+        // expect atleast one opportunity exists
```
chungthuang, line 10: Nit: at least.

tests/opportunity.spec.ts
```diff
@@ -0,0 +1,41 @@
...
+        expect(opportunityData.length).not.toBe(0)
```
chungthuang, line 11: Should we assert there are 2?

init/config.js
```diff
@@ -61,4 +61,34 @@ module.exports = {
     calendarId: "c_9c8sc215f8vopmjuj4ogamd7go@group.calendar.google.com",
     discordWebhookUrl: null
   },
+  firstOpportunity: {
```
chungthuang, line 64: We can make this more flexible by defining opportunities as an array of opportunity.

tests/opportunity.spec.ts
```diff
@@ -0,0 +1,40 @@
...
+    test("Get all opportunities", async ({ request }) => {
+        const opportunityData = await api(request).getData("/api/opportunities");
+
+        // expect two opportunities to exists
+        expect(opportunityData.length).toBe(2)
+
+        const opportunity = opportunityData[0].attributes;
+        
+        // expect opportunities have different fields filled
+        expect(opportunity.title).toBe(config.firstOpportunity.title);
```
chungthuang, line 16: Should we also assert the second opportunity in this test?

init/common.js
```diff
@@ -202,6 +206,19 @@ const createProject = async (interestId, userId, project, addTeam=true) => {
   });
 };
 
+const createOpportunity = async (interestId, opportunity) => {
+    let existing = await strapi.entityService.findMany("api::opportunity.opportunity", {filters: {title: opportunity.title}});
+    existing = existing[0];
```
tiagopazhs, line 211: I checked here and if this:
`
await strapi.entityService.findMany("api::opportunity.opportunity", {filters: {title: opportunity.title}});
` is null. When you try to access that, may it return an error:

```
> existing = null
null
> existing[0]
Uncaught TypeError: Cannot read properties of null (reading '0')
```

init/common.js
```diff
@@ -202,6 +206,19 @@ const createProject = async (interestId, userId, project, addTeam=true) => {
...
+    if (existing) {
+      return;
```
tiagopazhs, line 213: I don't know very well how it works. This is a question, why is it necessary to scape if `existing` ? Should we trown an error in this case ?

init/common.js
```diff
@@ -202,6 +206,19 @@ const createProject = async (interestId, userId, project, addTeam=true) => {
...
+    }
+  
+    opportunity.interests = [interestId];
+    
+    const service = strapi.service("api::opportunity.opportunity");
+    existing = await service.create({ data: opportunity });
+};
```
tiagopazhs, line 220: As we are not using existing after line 219, we can replace this:
```existing = await service.create({ data: opportunity });```
to just this ?
```await service.create({ data: opportunity });```

**Review by tiagopazhs — CHANGES_REQUESTED** [review]

File: init/common.js (hunk 8 above)
Comment (nikita-pardeshi-github) [reply init/common.js]: @tiagopazhs good catch. Handled exception.

File: init/common.js (hunk 9 above)
Comment (nikita-pardeshi-github) [reply init/common.js:220]: We are trying to avoid creating opportunity if one with the same title already exists. We are just generating test data here so that all tests can use it if required.

**Comment by nikita-pardeshi-github** [comment]
@chungthuang @tiagopazhs I pushed my changes. Can you please review once?

**Comment by tiagopazhs** [comment]
Hey @nikita-pardeshi-github can you give me instructions on how to reproduce this changes ?

**Comment by dbradham** [comment]
@tiagopazhs can you review Nikita's latest changes here?

File: init/common.js [hunk 11]
```diff
@@ -202,6 +204,30 @@ const createProject = async (interestId, userId, project, addTeam=true) => {
...
+
+  for (let index = 0; index < config.opportunity.length; index++) {
+    try{
+      let existing = await strapi.entityService.findMany("api::opportunity.opportunity", {filters: {title: config.opportunity[index].title}});
+      existing = existing[0];
+
+      if (existing) {
+        return;
+      }
+
```
tiagopazhs, line 217: I have a question. Here do you want to stop just the current iteration. Or the entire loop ?
And why ?

tiagopazhs:
> Hey @nikita-pardeshi-github can you give me instructions on how to reproduce this changes ?

Hey @nikita-pardeshi-github as you suggested, running `npm run test:coverage` I could test that.
It is working as expected!

```
1 skipped
24 passed (10.6s)
```

init/common.js
```diff
@@ -61,6 +61,8 @@ const bootstrapDatabase = async () => {
   // creating project without associating team
   await createProject(interestId, userId, config.projectWithoutTeam, false);
   console.log("Project created without associated team");
+  await createOpportunities(interestId);
+  console.log("Opportunities created");
```
tiagopazhs, line 65: Hey @nikita-pardeshi-github I realize that here we are using plural. In this case we are creatting opportunities or just one opportunity right?

tests/opportunity.spec.ts
tiagopazhs: hey @nikita-pardeshi-github I think as you created `"/api/opportunities/2"` you can mark this as resolved.

tiagopazhs (commented):
Hey Nikita

init/common.js
nikita-pardeshi-github: @tiagopazhs we are creating more than one opportunities at this instance as mentioned in user config. Considering opportunities declared in user config we can keep it plural for now, would that be okay? Additionally, can you please approve if you do find it good to merge now?

tiagopazhs (approved):

dbradham (approved):
looks good, thanks for responding to these comments!

PR #214, opened 2024-08-17
feat(notifications): Expose a PUT endpoint to update notification

chungthuang:
This should only be used to update a notification readDataeTime


tests/notification.spec.ts
chungthuang, line 1: @nikita-pardeshi-github can you help complete the other tests cases for the new update endpoint? We need to make sure we have the policy controller working correctly, such that it only allows you to update notifications that you own.

oriyomibadmus (approved):

PR #217, opened 2024-08-22
change the types of objectIds for likes

dbradham:


chungthuang (approved):

src/api/like/content-types/like/schema.json
```diff
@@ -17,15 +17,15 @@
       "relation": "oneToOne",
       "target": "plugin::users-permissions.user"
     },
-    "objectId": {
-      "type": "uid"
```
chungthuang, line 21: Oh we didn't catch this during code review.
@nikita-pardeshi-github let's also test the like collection.

PR #220, opened 2024-08-31
package.json in the type folder

VictorDiniz89:
I created a script "install-all" pointing for the path and right now we have a package.json in the type folder

### DIFF
--- a/package.json
+++ b/package.json
@@ -13,8 +13,7 @@
     "strapi": "strapi",
     "test": "playwright test",
     "test:postgres": "TEST_CONTAINER=true nyc playwright test",
-    "test:coverage": "nyc playwright test",
-     "install-all": "npm install --prefix /home/runner/work/strapiv4/strapiv4/"
+    "test:coverage": "nyc playwright test"
   },
   "dependencies": {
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -13,7 +13,8 @@
     "strapi": "strapi",
     "test": "playwright test",
     "test:postgres": "TEST_CONTAINER=true nyc playwright test",
-    "test:coverage": "nyc playwright test"
+    "test:coverage": "nyc playwright test",
+     "install-all": "npm install --prefix /home/runner/work/strapiv4/strapiv4/"
```
chungthuang, line 17: Where are we calling `install-all`? Note that https://github.com/dev-launchers/strapiv4/blob/main/.github/workflows/publish-types.yaml#L26 still goes into the type folder. Do we need to go into the types folder?

VictorDiniz89: I removed install-all and it didn't change notting inside the code working.
We don't need to go to type folder for running our npm publish
We can do it by '...your path/strapiv4' or '...your path/strapiv4/types'

VictorDiniz89:
Finished all switches for this PR

chungthuang (approved):

PR #222, opened 2024-09-08
feat(Profile): #1930 update profile schema and tested

anne-creator:
add Location, Role, Headline, Website as text under Profile Schema, and tested.

### DIFF
--- a/init/common.js
+++ b/init/common.js
@@ -61,6 +61,8 @@ const bootstrapDatabase = async () => {
   // creating project without associating team
   await createProject(interestId, userId, config.projectWithoutTeam, false);
   console.log("Project created without associated team");
+  await createProfile();
+  console.log("Profile created");
   return instance;
 };
 
@@ -202,6 +204,11 @@ const createProject = async (interestId, userId, project, addTeam=true) => {
   });
 };
 
+const createProfile = async () => {
+  const service = strapi.service("api::profile.profile");
+  const newProfile = await service.create({ data: config.profile });
+};
+
 module.exports = {
   setupEnv,
   setupStrapi,
@@ -215,4 +222,5 @@ module.exports = {
   setupGoogleAuthProvider,
   createInterests,
   createProject,
+  createProfile,
 };
--- a/src/api/profile/content-types/profile/schema.json
+++ b/src/api/profile/content-types/profile/schema.json
@@ -4,7 +4,7 @@
   "info": {
     "singularName": "profile",
     "pluralName": "profiles",
-    "displayName": "profile",
+    "displayName": "Profile",
     "description": ""
   },
   "options": {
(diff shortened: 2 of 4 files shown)


src/api/profile/content-types/profile/schema.json
```diff
@@ -44,6 +44,18 @@
       "config": {
         "provider": "azure-storage"
       }
+    },
+    "location": {
```
chungthuang, line 48: Maybe we can give it a more specific name? Is location a zip code, country, time zone or something else?

src/api/profile/content-types/profile/schema.json
```diff
@@ -44,6 +44,18 @@
...
+      "type": "string"
+    },
+    "role": {
```
chungthuang, line 51: Is this the user's role (developer, designer, etc) inside Dev Launchers?

src/api/profile/content-types/profile/schema.json
```diff
@@ -44,6 +44,18 @@
...
+      "type": "string"
+    },
+    "headline": {
+      "type": "string"
+    },
+    "website": {
```
chungthuang, line 57: What kind of website will this be? Someone's portfolio, LinkedIn, etc? Maybe this can be a list?

src/api/profile/content-types/profile/schema.json
judeamos: Hey @chungthuang, this will describe how they want to do or are doing - a designer, developer etc.. 
Basically how they want to be viewed as.  I've attached a screenshot of it

image

src/api/profile/content-types/profile/schema.json
judeamos: This will be a portfolio website. We can add a separate linkedin later.

src/api/profile/content-types/profile/schema.json
chungthuang: So this looks like we have category of roles (developer, UX/UI, other) which is similar to [roleCategory in opportunity collection](https://github.com/dev-launchers/strapiv4/blob/main/src/api/opportunity/content-types/opportunity/schema.json#L59). I think this is a good use case for components, so we share the same role category across our collections.

src/api/profile/content-types/profile/schema.json
chungthuang: Maybe we can call this field `portfolioWebsite`.

src/api/profile/content-types/profile/schema.json
chungthuang: Something like https://github.com/dev-launchers/strapiv4/commit/64f20421b0880f413b71365fe5ad5232fd6295d9.

KaitlynMallare:
@chungthuang Is this a valid story or has it already been resolved?

init/common.js
```diff
@@ -202,6 +204,11 @@ const createProject = async (interestId, userId, project, addTeam=true) => {
   });
 };
 
+const createProfile = async () => {
+  const service = strapi.service("api::profile.profile");
+  const newProfile = await service.create({ data: config.profile });
+};
+
```
j-peace, line 211: Is this newProfile associated to the current user?
How it works?

j-peace (changes requested):

PR #224, opened 2024-09-19
Testing publish workflow

VictorDiniz89:
I made some changes in the code workflow/publish-types for solving some errors log @chungthuang

### DIFF
--- a/.github/workflows/publish-types.yaml
+++ b/.github/workflows/publish-types.yaml
@@ -7,6 +7,14 @@ on:
     branches:
       - main
 
+    paths-ignore:
+      # Ignore when changes is made by flux
+      - 'staging/**'
+      - 'production/**'
+      # Ignore markdown files
+      - "**/*.md"
+    # Allows you to run this workflow manually from the Actions tab
+
 jobs:
   publish:
     runs-on: ubuntu-latest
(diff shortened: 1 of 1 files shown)


.github/workflows/publish-types.yaml
chungthuang, line 1: I think we should remove https://github.com/dev-launchers/strapiv4/pull/224/files#diff-5004bbe418c007bef8a1a3726eedc7f777fc066ae6279129670b3b2abda69a92R4-R5, because it's trying to publish when I create a release PR https://github.com/dev-launchers/strapiv4/pull/225/checks.

.github/workflows/publish-types.yaml
```diff
@@ -18,6 +26,10 @@ jobs:
         uses: actions/setup-node@v2
         with:
           node-version: '20'
+          registry-url : 'https://registry.npmjs.org'
+
+      - name: Configure npm
+        run: echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > ~/.npmrc
```
chungthuang, line 32: Should we add this to https://github.com/dev-launchers/strapiv4/blob/main/.npmrc directly?

VictorDiniz89: yes, we should, adding this line

      - name: Configure npm
        run: |
          echo "//registry.npmjs.org/:_authToken=${{ secrets.NPM_TOKEN }}" > ~/.npmrc
          echo "registry=https://registry.npmjs.org" >> ~/.npmrc

.github/workflows/publish-types.yaml
VictorDiniz89: ok, I did it

name: Publish Types

on:

  push:
    branches:
      - main

chungthuang (approved):

PR #226, opened 2024-09-23
feat(event): Record the user that initiated a notification event

oriyomibadmus:


dbradham (approved):
Looks good, thanks for making minimal changes!

PR #228, opened 2024-09-24
feat: Create a reusable component for role category and migrate opportunity to use it

### LINKED ISSUE #227: [FEAT] Create a reusable component for role category
**Summary**
The opportunity and [profile collection](https://github.com/dev-launchers/strapiv4/pull/222#discussion_r1755698798)s both want to collect role category. We should have a shared definition.

**Describe the solution you'd like**
Create a reusable component for role category that can be referenced by the opportunity and profile collections.

**Acceptance criteria**
We already have integration test for the opportunity collection. It should continue to pass after this change.

**Additional context**
We will migrate staging and production data for opportunity collection manually.

chungthuang:
Resolves #227

### DIFF
--- a/init/config.js
+++ b/init/config.js
@@ -61,7 +61,7 @@ module.exports = {
     calendarId: "c_9c8sc215f8vopmjuj4ogamd7go@group.calendar.google.com",
     discordWebhookUrl: null
   },
-  opportunity : [{
+  opportunity: [{
     title: "test opportunity",
     level: "Beginner",
     commitmentHoursPerWeek: 2,
@@ -74,7 +74,12 @@ module.exports = {
     ],
     isHidden: true,
     roleCategory: "Development",
-    roleType: "Back-End Developer"
+    roleType: "Back-End Developer",
+    role: {
+      "id": 1,
+      "name": "Back-End Developer",
+      "category": "Development"
+    }
   },
   {
     title: "test second opportunity",
@@ -88,7 +93,12 @@ module.exports = {
       }
     ],
     isHidden: true,
-    roleCategory: "Development",
-    roleType: "Back-End Developer"
+    roleCategory: "Product / UX",
+    roleType: "Product Lead",
+    role: {
+      "id": 2,
+      "name": "Product Lead",
+      "category": "Product / UX"
+    }
   }]
 };
--- a/src/api/opportunity/content-types/opportunity/schema.json
+++ b/src/api/opportunity/content-types/opportunity/schema.json
@@ -84,6 +84,11 @@
       ],
       "required": true,
       "default": "Front-End Developer"
+    },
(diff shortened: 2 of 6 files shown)


src/api/opportunity/content-types/opportunity/schema.json
```diff
@@ -84,6 +73,11 @@
       ],
       "required": true,
       "default": "Front-End Developer"
+    },
+    "roleCategory": {
+      "type": "component",
+      "repeatable": false,
+      "component": "role-category.role-category"
```
tiagopazhs, line 80: What this repeatable means ?

chungthuang: Repeatable means make this column a list. That would make roleCategory a list of values which wouldn't make sense, because each opportunity belongs to one category.

src/components/role-category/role-category.json
```diff
@@ -0,0 +1,20 @@
...
+      "enum": [
+        "Product / UX",
+        "Development",
+        "QA",
+        "Operations",
+        "Other"
+      ]
+    }
+  }
+}
```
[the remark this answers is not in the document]
chungthuang: `roleType` is the previous column used by the opportunity collection that will be replaced by the roleCategory. My plan is:
1. Merge and deploy roleCategory
2. Set roleCategory to be the same value as roleType
3. Delete roleType
Ideally we can write a migration to automatically set roleCategory to the same value as roleType. However, given that we only have a few opportunities in the database, it's faster to do the migration manually.

src/api/opportunity/content-types/opportunity/schema.json
tiagopazhs: OK, thank you, it is clear now.

src/components/role-category/role-category.json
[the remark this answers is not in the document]
tiagopazhs: Hmm maybe it is something like a subcategory ? 

Let me know if it is correct. 
We can create a model roleCategory that has the attribute category.
But to make it practical we can use it as a component ?

It is possible to cre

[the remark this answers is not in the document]
chungthuang: It's not a subcategory, but a new type. Under this new type, there is an attribute called category.
We make this type into a component, so it can be reused between collections.

init/config.js
```diff
@@ -61,7 +61,7 @@ module.exports = {
     calendarId: "c_9c8sc215f8vopmjuj4ogamd7go@group.calendar.google.com",
     discordWebhookUrl: null
   },
-  opportunity : [{
+  opportunity: [{
```
dbradham, line 64: what a great opportunity to clean code!

dbradham (approved):
looks good!

PR #230, opened 2024-09-29
Tiago/51 cannot reset password with admin endpoint

tiagopazhs:
## Changes
- Adding the pluging to enable send emails using strapi.

## How to test
- Generate a gmail passkey
- Go to config/plugins.js, and include your email and passkey at email.config.providerOptions.auth
- Go to http://localhost:1337/admin/settings/email and send and email test.

## Next steps
- Move the credentials to a .env file.
- Enable the route forgot-pasword.

### DIFF
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -64,5 +64,22 @@ module.exports = ({ env }) => ({
         name: "strapi-prometheus",
       },
     }
-  }
+  },
+  email: {
+    config: {
+      provider: 'nodemailer',
+      providerOptions: {
+        host: 'smtp.gmail.com',
+        port: 465,
+        auth: {
+          user: "email@gmail.com", // edit here
+          pass: "zzzz zzzz zzzz zzzz", // edit here
+        },
+      },
+      settings: {
+        defaultFrom: 'hello@example.com',
+        defaultReplyTo: 'hello@example.com',
+      },
+    },
+  },
 });
--- a/config/plugins.js
+++ b/config/plugins.js
@@ -72,13 +72,13 @@ module.exports = ({ env }) => ({
         host: 'smtp.gmail.com',
         port: 465,
         auth: {
-          user: "email@gmail.com", // edit here
-          pass: "zzzz zzzz zzzz zzzz", // edit here
+          user: env('NODE_MAILER_EMAIL'),
+          pass: env('NODE_MAILER_PASSWORD')
         },
       },
       settings: {
-        defaultFrom: 'hello@example.com',
-        defaultReplyTo: 'hello@example.com',
+        defaultFrom: env('NODE_MAILER_EMAIL'),
+        defaultReplyTo: env('NODE_MAILER_EMAIL'),
       },
     },
   },
--- a/package-lock.json
+++ b/package-lock.json
@@ -13,6 +13,7 @@
         "@strapi/plugin-i18n": "4.15.0",
         "@strapi/plugin-sentry": "4.15.0",
         "@strapi/plugin-users-permissions": "4.15.0",
+        "@strapi/provider-email-nodemailer": "^5.0.1",
         "@strapi/strapi": "4.15.0",
         "better-sqlite3": "^9.2.2",
         "discord-webhook-node": "^1.1.8",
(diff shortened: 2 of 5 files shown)


config/plugins.js
```diff
@@ -64,5 +64,22 @@ module.exports = ({ env }) => ({
...
+  },
+  email: {
+    config: {
+      provider: 'nodemailer',
+      providerOptions: {
+        host: 'smtp.gmail.com',
+        port: 465,
+        auth: {
+          user: "email@gmail.com", // edit here
+          pass: "zzzz zzzz zzzz zzzz", // edit here
```
dbradham, line 76: @chungthuang @oriyomibadmus how could we add an environment variable that is accessible in the staging or production environments? Tiago got this working using my credentials to test with, but ideally we could set up the ability to have dev launchers credentials which are not exposed in the source control.

config/plugins.js
```diff
@@ -64,5 +64,22 @@ module.exports = ({ env }) => ({
...
+        },
+      },
+      settings: {
+        defaultFrom: 'hello@example.com',
```
dbradham, line 80: @chungthuang @Enjoy2Live @annaisaacks @leeah-d is there an email address we want to use as the "from" address by default for dev launchers email interactions? We want this to be something professional because users may encounter emails during their onboarding and ongoing contributions at dev launchers.

For example, at my company, we have a "support@" email that is broadcasted to our operations and central support teams.

config/plugins.js
chungthuang: Our standard is adding a secret to the kuberentes server by encrpyting it and add to https://github.com/dev-launchers/k8s-infra/tree/main/tenants/production/strapiv4/secrets. The app can access it via an environmental variable, for example, https://github.com/dev-launchers/strapiv4/blob/main/workload/deployment.yaml#L50-L54.

config/plugins.js
chungthuang: We have support@devlaunchers.com.

config/plugins.js
tiagopazhs: Something like:

```
./workload/deployment.yaml

- name: CLIENT_SECRET # replace for 'USER' or/and 'PASSWORD'
    valueFrom:
      secretKeyRef:
        name: auth # replace for 'nodemailer'
        key: discord-client-secret # replace for 'user' or/and 'password'
```

?

chungthuang: @tiagopazhs yes we need to add what you said to ./workload/deployment.yaml. And here we can reference the secret through environmental variable by `process.env(USER)`. I would recommending calling it something more specific, like `NODE_MAILER_USER`.

tiagopazhs: > can reference the secret through environmental variable by process.env(USER).
OK

> calling it something more specific, like NODE_MAILER_USER.
Good! I appreciate this suggestion!

dbradham (approved):
Looks good to me!

dbradham:
@chungthuang I think we figured this out, but wanted to ask if you could take a look when you get a chance. We mainly want confirmation that the `deployment.yaml` looks as you would except here.

chungthuang (approved):

chungthuang:
The deployment file looks correct!

tiagopazhs:
Draft of Enviroment Variables.

I am planning to record this information bellow this section (development process): https://github.com/dev-launchers/dev-launchers-platform/wiki#development-processes.
Please tell me if you can se a better place to save that


## Development Processes

### Use enviroment variables

- How to use in staging and development.
* It is not possible to use the same scenario of enviroment variables in staging and development, so to test a feature is necessary to mock your secret (don't push the secrets to a remote branch).

- How to use in production.
* Follow the step by step on how to set a variable to use in production:

#### Step 1: Path to storage the secret

- Save the secret inside the repository https://github.com/dev-launchers/k8s-infra.
- Having acess to the repository, save the secret on tenants/base/{project that will use the secret}/screts.

#### Step 2: Encrypting the secret

- {Adding here how to encrypting the secret}
- Follow the following structure to record a secret: https://github.com/dev-launchers/k8s-infra/blob/main/tenants/base/strapiv4/secrets/nodemailer.yaml (here the secrets are `user` and `password` under `stringData`

PR #232, opened 2024-10-06
feat(event): Update event schema with action and (Latest) entityName, removed title

oriyomibadmus:
The event schema has been updated with the **action** that triggered its creation, the latest **entityName** which references the collection that created it (e.g: ideaName, CommentText, etc). The **title** is removed to enable easy consumption of the API.

### DIFF
--- a/src/api/event/content-types/event/schema.json
+++ b/src/api/event/content-types/event/schema.json
@@ -21,9 +21,6 @@
         "IdeaCard"
       ]
     },
-    "title": {
-      "type": "string"
-    },
     "content": {
       "type": "text"
     },
@@ -34,6 +31,18 @@
       "type": "relation",
       "relation": "oneToOne",
       "target": "plugin::users-permissions.user"
+    },
+    "action": {
+      "type": "enumeration",
+      "enum": [
+        "Idea Created",
+        "Idea Updated",
+        "Commented",
+        "Liked"
+      ]
+    },
+    "entityName": {
+      "type": "string"
     }
   }
 }
--- a/src/api/event/content-types/event/schema.json
+++ b/src/api/event/content-types/event/schema.json
@@ -18,7 +18,8 @@
     "entityType": {
       "type": "enumeration",
       "enum": [
-        "IdeaCard"
+        "IdeaCard",
+        "Comment"
       ]
     },
     "content": {
--- a/src/api/event/controllers/event.js
+++ b/src/api/event/controllers/event.js
@@ -1,9 +1,30 @@
-'use strict';
-
-/**
- * event controller
- */
-
 const { createCoreController } = require('@strapi/strapi').factories;
 
-module.exports = createCoreController('api::event.event');
+module.exports = createCoreController('api::event.event', ({ strapi }) => ({
+
+  async find(ctx) {
+    const events = await strapi.entityService.findMany('api::event.event', {
+      populate: ['eventUser'],
+    });
+
+    // Retrieve latest entityName for each event Dynamically
+    const enhancedEvents = await Promise.all(events.map(async (event) => {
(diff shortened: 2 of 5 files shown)


oriyomibadmus:
Screenshot 2024-10-06 at 6 48 54 PM
Screenshot 2024-10-06 at 6 49 17 PM
Screenshot 2024-10-06 at 6 49 08 PM

dbradham (approved):
Looks good!

src/api/event/content-types/event/schema.json
```diff
@@ -34,6 +32,18 @@
       "type": "relation",
       "relation": "oneToOne",
       "target": "plugin::users-permissions.user"
+    },
+    "action": {
```
chungthuang, line 36: I think we should consider `entityType` + `action` when we define the enum values. What does `entityType` Comment with `action` Idea Created mean? It probably doesn't make sense, but it's a valid combination.
I don't have a better suggestion right now. Each `entityType` has its set of actions.

src/api/event/controllers/event.js
```diff
@@ -1,9 +1,38 @@
-'use strict';
+const { createCoreController } = require('@strapi/strapi').factories;
 
-/**
- * event controller
- */
+module.exports = createCoreController('api::event.event', ({ strapi }) => ({
+  async find(ctx) {
```
chungthuang, line 4: Does this mean we only overwrite the find method? Other methods like update are still using strapi's default implementation?

chungthuang, line 4: Is this controller shared between finding a specific event and listing all events?

src/api/event/controllers/event.js
```diff
@@ -1,9 +1,38 @@
...
+    const latestEvents = await Promise.all(events.map(async (event) => {
+      let entityName = '';
+
+      if (event.entityType === 'IdeaCard') {
+        const ideaCard = await strapi.entityService.findOne('api::idea-card.idea-card', event.entityId);
+        entityName = ideaCard ? ideaCard.ideaName : 'Unknown Idea';
+      } else if (event.entityType === 'Comment') {
+        const comment = await strapi.entityService.findOne('api::comment.comment', event.entityId);
+        entityName = comment ? comment.text : 'Unknown Comment';
+      }
```
chungthuang, line 28: Should we have an else case to return an error for unknown entityType?

src/api/event/controllers/event.js
```diff
@@ -1,9 +1,38 @@
...
+      populate: ['eventUser'],  // Populate relations (e.g., eventUser)
+    });
+
+    // Retrieve latest entityName for each event Dynamically
+    const latestEvents = await Promise.all(events.map(async (event) => {
+      let entityName = '';
+
+      if (event.entityType === 'IdeaCard') {
+        const ideaCard = await strapi.entityService.findOne('api::idea-card.idea-card', event.entityId);
+        entityName = ideaCard ? ideaCard.ideaName : 'Unknown Idea';
```
chungthuang, line 24: Should we return an error if we can't find the idea/comment? Alternatively, we can drop the event and log an error on the server side so we know something unexpected happened (the idea/comment is missing).

src/api/event/controllers/event.js
```diff
@@ -1,9 +1,38 @@
...
+      } else if (event.entityType === 'Comment') {
```
tiagopazhs, line 25: Hey @oriyomibadmus just a minor comment here.

This `else` is necessary ? Can we remove that and get the same result ?

```
else if (event.entityType === 'Comment')
```

Now, in order to catch other scenarios, another question, `event.entityType` can be != 'IdeaCard' and != 'Comment' ?

PR #233, opened 2024-10-12
Oriyomibadmus/test update

oriyomibadmus:
**Update Highlights:**

1. **Enhanced Playwright Tests:** Adjusted the Playwright test suite to align with the updated `event` schema, ensuring correct validation and handling of notifications and events.
2. **Refined Comment Lifecycle**: Modified the comment lifecycle to use the `commentId` as the `entityId` in the event model, instead of the `ideaCardId`, allowing more precise tracking of comment-related events.

### DIFF
--- a/src/api/event/controllers/event.js
+++ b/src/api/event/controllers/event.js
@@ -1,9 +1,30 @@
-'use strict';
-
-/**
- * event controller
- */
-
 const { createCoreController } = require('@strapi/strapi').factories;
 
-module.exports = createCoreController('api::event.event');
+module.exports = createCoreController('api::event.event', ({ strapi }) => ({
+
+  async find(ctx) {
+    const events = await strapi.entityService.findMany('api::event.event', {
+      populate: ['eventUser'],
+    });
+
+    // Retrieve latest entityName for each event Dynamically
+    const enhancedEvents = await Promise.all(events.map(async (event) => {
+      let entityName = '';
+
+      if (event.entityType === 'IdeaCard') {
+        const ideaCard = await strapi.entityService.findOne('api::idea-card.idea-card', event.entityId);
+        entityName = ideaCard ? ideaCard.ideaName : 'Unknown Idea';
+      } else if (event.entityType === 'Comment') {
+        const comment = await strapi.entityService.findOne('api::comment.comment', event.entityId);
+        entityName = comment ? comment.text : 'Unknown Comment';
+      }
+
+      return {
+        ...event,
+        entityName,  // Adding the dynamic (latest) entityName to the event
+      };
+    }));
+
+    return enhancedEvents;
+  },
+}));
--- a/src/api/event/controllers/event.js
+++ b/src/api/event/controllers/event.js
@@ -1,14 +1,22 @@
 const { createCoreController } = require('@strapi/strapi').factories;
 
 module.exports = createCoreController('api::event.event', ({ strapi }) => ({
-
   async find(ctx) {
+    // filter with the query
+    const { entityId, entityType } = ctx.query;
+
+    // Fetch only the events that match the filters
+    const filters = {};
+    if (entityId) filters.entityId = entityId;
+    if (entityType) filters.entityType = entityType;
+
     const events = await strapi.entityService.findMany('api::event.event', {
-      populate: ['eventUser'],
+      filters,  // Apply filters to fetch specific events
+      populate: ['eventUser'],  // Populate relations (e.g., eventUser)
     });
 
     // Retrieve latest entityName for each event Dynamically
-    const enhancedEvents = await Promise.all(events.map(async (event) => {
+    const latestEvents = await Promise.all(events.map(async (event) => {
       let entityName = '';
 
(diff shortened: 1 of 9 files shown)


src/api/event/controllers/event.js
```diff
@@ -1,9 +1,38 @@
...
+    if (entityId) filters.entityId = entityId;
+    if (entityType) filters.entityType = entityType;
+
+    const events = await strapi.entityService.findMany('api::event.event', {
+      filters,  // Apply filters to fetch specific events
+      populate: ['eventUser'],  // Populate relations (e.g., eventUser)
+    });
+
+    // Retrieve latest entityName for each event Dynamically
+    const latestEvents = await Promise.all(events.map(async (event) => {
```
tiagopazhs, line 19: Hey Oriyomi, Promisse.all braks my mind a litle bit.
Do you know the advantage to use that here in this case ?

src/api/event/controllers/event.js
```diff
@@ -1,9 +1,38 @@
...
+      let entityName = '';
+
+      if (event.entityType === 'IdeaCard') {
+        const ideaCard = await strapi.entityService.findOne('api::idea-card.idea-card', event.entityId);
+        entityName = ideaCard ? ideaCard.ideaName : 'Unknown Idea';
+      } else if (event.entityType === 'Comment') {
+        const comment = await strapi.entityService.findOne('api::comment.comment', event.entityId);
+        entityName = comment ? comment.text : 'Unknown Comment';
+      }
```
chungthuang, line 28: Can we add an else case to catch unexpected `entityType`? This can happen if we add a new type to the schema, but forget to update this logic.

tests/event.spec.ts
```diff
@@ -18,47 +19,85 @@ test.describe('/api/event', () => {
             involveLevel: "minimum",
             status: "workshopping",
         });
+        // Get the Idea Id for testing
         ideaId = newIdea.id;
+        // Get all the events
         const events = await api(request).get("/api/events");
```
chungthuang, line 25: Does the endpoint support query parameters? I think it should, and we can filter by passing `?entityType=IdeaCard`.

tests/event.spec.ts
```diff
@@ -18,47 +19,85 @@ test.describe('/api/event', () => {
...
-        expect(event.attributes.content).toBe(`${config.user.username} has updated idea TestEventIdea`);
+        // Get all the events
+        const events = await api(request).get('/api/events');
+        // Filter the events for the one related to the recently updated Idea
+        const event = events.reverse().find(item => (item.entityId === idea.id && item.entityType === 'IdeaCard'));
+
+        // Schema Checks
+        expect(event).toBeDefined(); // Check that the event exists
+        expect(event.action).toBe('Idea Updated'); // Check the action
+        expect(event.entityName).toBe('TestEventIdea'); // Check the entity name
```
chungthuang, line 55: We should also validate the entityType.

tests/event.spec.ts
```diff
@@ -18,47 +19,85 @@ test.describe('/api/event', () => {
...
+
+        // Fetch all events
         const events = await api(request).get("/api/events");
-        const event = events.data.reverse()
-        .find(item => (item.attributes.entityId === idea.id && item.attributes.entityType === "IdeaCard"));
-        expect(event.attributes.title).toBe("Idea TestEventIdea is approved");
-        expect(event.attributes.content).toBe(`${config.user.username} has updated idea TestEventIdea - approved`);
+
+        // Find the event related to the updated idea status
+        const event = events.reverse().find(item => (item.entityId === updatedIdea.id && item.entityType === 'IdeaCard'));
```
chungthuang, line 70: Why reverse?

tests/notification.spec.ts
```diff
@@ -20,38 +21,62 @@ test.describe('/api/notification', () => {
...
-        expect(notification.attributes.event.data.attributes.content).toBe('test comment');
-        expect(notification.attributes.user.data.attributes.username).toBe(config.user.username);
-        expect(notification.attributes.createdDateTime).not.toBeNull();
+        const newCommentId = newComment.id;
+
+        // Fetch notifications after the comment is created
+        const notifications = await api(request).get("/api/notifications?populate=event,user");
+
+        // Filter the notification related to the comment using the comment ID
+        const notification = notifications.data.reverse().find(
```
chungthuang, line 60: Why reverse?

src/api/event/controllers/event.js
dbradham: The Promise.all allows us to run all of the processing in parallel. When we call `.map` the list is processed in sequence. Handling all of the events in the `.all()` allows us to run this process in parallel.

src/api/event/controllers/event.js
dbradham: Because this code is not authored by Yomi, we are approving these changes and making a follow up ticket for your relevant comments here. Thanks!

dbradham (approved):
Looks good to me. We discussed these changes on a couple of our weekly calls as a group.

PR #248, opened 2024-11-25
fix(notification) : user comment subscription for notification

yacineKahlerras:
- user that comments is subscribed to the idea card not the comment
- when user comments all who are subscribed to the idea card get a notification

### DIFF
--- a/src/api/comment/content-types/comment/lifecycles.js
+++ b/src/api/comment/content-types/comment/lifecycles.js
@@ -19,15 +19,25 @@ module.exports = {
     const idea = comment.idea_card;
     const user = comment.user;
 
-    await strapi.entityService.create("api::subscription.subscription", {
-      data: {
-        entityType: "Comment",
-        entityId: commentId,
-        createdDateTime: new Date(),
-        active: true,
-        user: user,
-      },
-    });
+    /** verify that user isn't already subscribed to ideaCard */
+    const existingSubscriptions = await strapi.entityService.findMany(
+      "api::subscription.subscription",
+      {
+        filters: { user: user, entityId: idea.id },
+        limit: 1,
+      }
+    );
+
+    if (!existingSubscriptions.length)
+      await strapi.entityService.create("api::subscription.subscription", {
+        data: {
+          entityType: "Comment",
+          entityId: idea.id,
+          createdDateTime: new Date(),
+          active: true,
+          user: user,
+        },
+      });
 
     strapi.entityService.create("api::event.event", {
       data: {
@@ -35,7 +45,7 @@ module.exports = {
         entityName: idea.ideaName,
         content: commentText,
         entityType: "Comment",
-        entityId: commentId,
+        entityId: idea.id,
         eventUser: user?.id,
         createdDateTime: new Date(),
       },
--- a/tests/event.spec.ts
+++ b/tests/event.spec.ts
@@ -1,101 +1,111 @@
-import { test, expect } from '@playwright/test';
-import { api } from './utils';
-import * as config from '../init/config';
+import { test, expect } from "@playwright/test";
+import { api } from "./utils";
+import * as config from "../init/config";
 
 let ideaId = 0;
 
-test.describe('/api/event', () => {
+test.describe("/api/event", () => {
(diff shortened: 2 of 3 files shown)


src/api/comment/content-types/comment/lifecycles.js
```diff
@@ -19,23 +19,32 @@ module.exports = {
...
-      data: {
-        entityType: "Comment",
-        entityId: commentId,
-        createdDateTime: new Date(),
-        active: true,
-        user: user,
-      },
-    });
+    /** verify that user isn't already subscribed to ideaCard */
+    const existingSubscriptions = await strapi.entityService.findMany(
```
chungthuang, line 23: Do we need to find all of them? Or is `findOne` enough?

src/api/comment/content-types/comment/lifecycles.js
```diff
@@ -19,23 +19,32 @@ module.exports = {
...
+      });
 
     strapi.entityService.create("api::event.event", {
       data: {
         action: "Commented",
         entityName: idea.ideaName,
         content: commentText,
         entityType: "Comment",
-        entityId: commentId,
+        entityId: idea.id,
```
chungthuang, line 47: Can you add an integration test that would catch this error?

chungthuang:
Can you update the commit message to follow semantic release format https://github.com/semantic-release/semantic-release#commit-message-format? For this PR, it will start with something like `fix(notification)`.

src/api/comment/content-types/comment/lifecycles.js
yacineKahlerras: yes one is enough, i'll limit it to one entry

src/api/comment/content-types/comment/lifecycles.js
yacineKahlerras: okay

yacineKahlerras:
@chungthuang updated the code as requested

src/api/comment/content-types/comment/lifecycles.js
chungthuang: Yes.

chungthuang:
@yacineKahlerras don't forget to fix the commit message and squash the commits.

yacineKahlerras:
@chungthuang done✔

tests/event.spec.ts
```diff
@@ -1,101 +1,111 @@
...
+      );
+
+    // Schema Checks
+    expect(event).toBeDefined(); // Ensure the event exists
+    expect(event.action).toBe("Commented"); // Check the action is "Commented"
+    expect(event.entityName).toBe("This is a test comment"); // The entityName should be the comment text
+    expect(event.content).toBe("This is a test comment"); // Check the event content
+    expect(event.entityType).toBe("Comment"); // Ensure the entityType is "Comment"
+    expect(event.entityId).toBe(ideaId); // The entityId should be the ideaId related to the comment
+    expect(event.eventUser.username).toBe(config.user.username); // Check the eventUser is the comment author
```
chungthuang, line 109: Will there also be an event for the idea owner, to notify someone commented on their idea?

chungthuang (approved):

tests/event.spec.ts
yacineKahlerras: @chungthuang no new event is created for the idea owner, when this **comment event** is created, the user and all who are already _subscribed_ to the idea (idea owner, users who commented) get notified

chungthuang: Ok that makes sense. Can this test verify that both the idea owner and commenter get notified?

chungthuang (approved):

tests/event.spec.ts
yacineKahlerras: @chungthuang no because the comment is made by the idea owner and he cant get notified multiple time for the same event

chungthuang: That sounds like we should use a different user to comment to make this test more realistic?

yacineKahlerras: yes we can login with **user2** and comment and then fetch the notifications resulting from that event

PR #249, opened 2024-12-01
fix(package): Fix the package name for published types

chungthuang:
https://github.com/dev-launchers/strapiv4/actions/workflows/publish-types.yaml is failing to authenticate. I think this will fix the authentication issue, but the publication won't work yet, because we are not updating the package version.


dbradham:
Sagar picked up this work

PR #255, opened 2024-12-29
fix(event): Record the ID of the idea/comment that generated an event

chungthuang:


src/api/comment/content-types/comment/lifecycles.js
```diff
@@ -33,19 +33,21 @@ module.exports = {
         data: {
           entityType: "Comment",
           entityId: idea.id,
+          originatedEntityId: commentId,
```
yacineKahlerras, line 36: do we need this field in subscriptions as well ?

yacineKahlerras (approved):

dbradham (approved):

src/api/comment/content-types/comment/lifecycles.js
chungthuang: Oh you are right, this field is not in the subscription.

PR #250, opened 2024-12-07
Strapi Plugin for Uploading Applicant Portfolios to Shared Google Drive

jeyanthi-sm:
A custom Strapi plugin that provides features for file management, including uploading a single file, deleting a file, and retrieving a file by its ID.

PR #258, opened 2025-01-06
The resume uploaded by the applicant to be stored in the backend for further processes in recruitment

jeyanthi-sm:
When the user uploads a file while applying for a role and click on submit, the file should be stored in STRAPI backend. The
Applicant collection type should reflect the file uploaded by each applicant.

https://github.com/dev-launchers/dev-launchers-platform/issues/1344


config/middlewares.js
```diff
@@ -78,11 +78,21 @@ module.exports = [
...
   },
+
   'strapi::poweredBy',
   'strapi::logger',
   'strapi::query',
-  'strapi::body',
   'strapi::favicon',
   'strapi::public',
   'plugin::users-permissions.jwt',
+  {
```
chungthuang, line 88: Looks like there is a conflict here because we recently added https://github.com/dev-launchers/strapiv4/blob/main/config/middlewares.js#L88-L94. W should keep that and the body middleware that you are adding.

config/plugins.js
```diff
@@ -63,6 +63,20 @@ module.exports = ({ env }) => ({
       customLabels: {
         name: "strapi-prometheus",
       },
-    }
-  }
-});
\ No newline at end of file
+    },
```
chungthuang, line 66: Seems like there is a conflict with the email plugin https://github.com/dev-launchers/strapiv4/blob/main/config/plugins.js#L68-L84.

chungthuang:
What is the difference between this PR and https://github.com/dev-launchers/strapiv4/pull/250/files?

jeyanthi-sm:
Added new fields, such as portfolioLink, portfolioFileId and extraInfo to the Applicant collection

src/api/applicant/content-types/applicant/schema.json
```diff
@@ -82,6 +82,15 @@
...
       "default": true,
       "required": true
+    },
+    "portfolioLink": {
+      "type": "string"
+    },
+    "portfolioFileId": {
+      "type": "string"
+    },
+    "extraInfo": {
```
chungthuang, line 92: What kind of info will we store?

jeyanthi-sm: Any additional information or concerns an applicant wishes to convey that do not fit under any of the other questions

jeyanthi-sm:
Closing this PR. The same change was added to another PR, and the changes have been merged into the main branch.

PR #260, opened 2025-01-12
Check if hasProfileAssociated when log in using Google

j-peace:
## Description

This solves the problem of duplicated profiles.

For more details of the problem check here: https://github.com/orgs/dev-launchers/projects/29/views/1?pane=issue&itemId=85712147&issue=dev-launchers%7Cstrapiv4%7C240

## Changes

Before create a profile we check this:
hasProfileAssociated ?
- no: create a profile with google displayName and PictureProfile
- yes: update displayName and PictureProfile (just if these are empty)

## How to test

1. Loging in using google.
2. To test the scenarios is necessary to edit or delete the profile associated using the strapi admin content manager

### DIFF
--- a/src/extensions/users-permissions/server/services/providers-registry.js
+++ b/src/extensions/users-permissions/server/services/providers-registry.js
@@ -95,19 +95,44 @@ const getInitialProviders = ({ purest }) => ({
       .auth(accessToken)
       .request();
 
-    const createdProfile = await strapi.query('api::profile.profile').create({
-      data: {
-        displayName: body.name,
-        profilePictureUrl: body.picture
-      }
+    const currentUser = await strapi.db.query('plugin::users-permissions.user').findOne({
+      where: { email: body.email }
     });
+
+    const associatedProfiles = await strapi.db.query('api::profile.profile').findMany({
+      where: { user: currentUser }
+    });
+
+    let currentProfile = ''
+    if (associatedProfiles.length) {
+      const profileAssociated = associatedProfiles[0]
+      if (!profileAssociated?.profilePictureUrl || !profileAssociated?.displayName) {
+        currentProfile = await strapi.db.query('api::profile.profile').update({
+          where: {user: currentUser.id},
+          data: {
+            displayName: profileAssociated?.displayName || body.name,
+            profilePictureUrl: profileAssociated?.profilePictureUrl || body.picture
+          }
+        });
+      }
+    } else {
+      currentProfile = await strapi.db.query('api::profile.profile').create({
+        data: {
+          displayName: body.name,
+          profilePictureUrl: body.picture,
+          user: currentUser.id,
+          publishedAt: new Date(),
+        }
+      });
+    }
+
 
     return {
       username: body.email.split("@")[0],
       email: body.email,
       googleId: body.sub,
       userId: uuidv4(),
-      profile: createdProfile.id,
+      profile: currentProfile.id,
     }
   },
   async github({ accessToken }) {


src/extensions/users-permissions/server/services/providers-registry.js
```diff
@@ -87,14 +87,21 @@ const getInitialProviders = ({ purest }) => ({
...
-  async google({ accessToken }) {
+  async google({ accessToken, ctx }) {
     const google = purest({ provider: 'google' });
 
     const { body } = await google
       .get("https://www.googleapis.com/oauth2/v3/userinfo")
       .auth(accessToken)
       .request();
 
+    const authenticatedUserId = ctx.state.user?.id;
```
dbradham, line 98: @oriyomibadmus

src/extensions/users-permissions/server/services/providers-registry.js
```diff
@@ -87,14 +87,21 @@ const getInitialProviders = ({ purest }) => ({
...
+
+    // TODO: Validation check if the profile is already created
+    // hasProfileAssociated ?
```
dbradham, line 101: let's query for a user with this email to determine if the user has already been created... we can get the email from `body.email`

src/extensions/users-permissions/server/services/providers-registry.js
```diff
@@ -87,14 +87,21 @@ const getInitialProviders = ({ purest }) => ({
         email: body.email,
       }));
   },
-  async google({ accessToken }) {
+  async google({ accessToken, ctx }) {
```
chungthuang, line 90: Are we sure `ctx` is passed into this method?

j-peace: Actually it was not working, I will remove that.

src/extensions/users-permissions/server/services/providers-registry.js
```diff
@@ -87,21 +87,41 @@ const getInitialProviders = ({ purest }) => ({
...
+    const currentUser = await strapi.db.query('plugin::users-permissions.user').findOne({
+      where: { email: body.email }
+    });
+
+    const hasProfileAssociated = await strapi.db.query('api::profile.profile').findMany({
+      where: { user: currentUser }
     });
 
+    if (hasProfileAssociated?.length > 0) {
+      const profileAssociated = hasProfileAssociated[0]
```
j-peace, line 107: @dbradham check if that bellow. If we don't have one of the properties, name or picture. We update the user with the one that is missed.

chungthuang: We already have users with multiple profiles created because of this bug. Should we clean up to make sure a user only has 1 profile?

src/extensions/users-permissions/server/services/providers-registry.js
```diff
@@ -87,21 +87,41 @@ const getInitialProviders = ({ purest }) => ({
...
+      if (profileAssociated?.profilePictureUrl || profileAssociated?.displayName) {
+        await strapi.query('api::profile.profile').update({
+          data: {
+            displayName: profileAssociated?.profilePictureUrl || body.name,
```
chungthuang, line 111: Why are we setting display name to `profilePictureUrl`?

src/extensions/users-permissions/server/services/providers-registry.js
```diff
@@ -87,21 +87,41 @@ const getInitialProviders = ({ purest }) => ({
...
+            profilePictureUrl: profileAssociated?.displayName || body.picture
```
chungthuang, line 112: I think we are mixing `profilePictureUrl` and `displayName`. You probably mean `profilePictureUrl` here.

types/generated/contentTypes.d.ts
```diff
@@ -362,98 +362,6 @@ export interface AdminTransferTokenPermission extends Schema.CollectionType {
   };
 }
 
-export interface AdminWorkflow extends Schema.CollectionType {
```
chungthuang, line 365: Why is this removed?

types/generated/contentTypes.d.ts
```diff
@@ -1667,47 +1575,6 @@ export interface ApiTeamMembershipTeamMembership extends Schema.CollectionType {
   };
 }
 
-export interface AdminAuditLog extends Schema.CollectionType {
```
chungthuang, line 1670: Why is this removed?

src/extensions/users-permissions/server/services/providers-registry.js
chungthuang: Yes, in strapi admin UI we can define the relationship to be 1-1 https://strapi.io/blog/strapi-relations-101.

src/extensions/users-permissions/server/services/providers-registry.js
chungthuang: `ctx` hasn't been removed.

src/extensions/users-permissions/server/services/providers-registry.js
j-peace: Actually, the relation was corret.
Just one to one.
But in the old scenario we were not associating profile with user (not passing the param):
image

src/extensions/users-permissions/server/services/providers-registry.js
j-peace: Done: 
image

src/extensions/users-permissions/server/services/providers-registry.js
j-peace: Yes, it was solved:
image

types/generated/contentTypes.d.ts
j-peace: Re-added

types/generated/contentTypes.d.ts
j-peace: re-added

dbradham (approved):

src/extensions/users-permissions/server/services/providers-registry.js
```diff
@@ -95,19 +95,43 @@ const getInitialProviders = ({ purest }) => ({
...
+    const hasProfileAssociated = await strapi.db.query('api::profile.profile').findMany({
+      where: { user: currentUser }
     });
 
+    let currentProfile = ''
+    if (hasProfileAssociated.length) {
+      const profileAssociated = hasProfileAssociated[0]
+      if (!profileAssociated?.profilePictureUrl || !profileAssociated?.displayName) {
+        currentProfile = await strapi.db.query('api::profile.profile').update({
+          where: {user: currentUser.id},
```
chungthuang, line 111: So we are updating all the profiles of the user? Should we delete the extra profiles?

src/extensions/users-permissions/server/services/providers-registry.js
```diff
@@ -95,19 +95,43 @@ const getInitialProviders = ({ purest }) => ({
...
-    const createdProfile = await strapi.query('api::profile.profile').create({
-      data: {
-        displayName: body.name,
-        profilePictureUrl: body.picture
-      }
+    const currentUser = await strapi.db.query('plugin::users-permissions.user').findOne({
+      where: { email: body.email }
+    });
+
+    const hasProfileAssociated = await strapi.db.query('api::profile.profile').findMany({
```
chungthuang, line 102: Nit: I would call this `assoicataedProfiles`. A variable starting with `has` usually indicate it's a boolean.

src/extensions/users-permissions/server/services/providers-registry.js
j-peace: Good question! We are updating just the first (and unique) profile. And atually this scenario of one user having more than one profile doesn't exists. We have a relation 1 to 1.

So what was the real problem? lol let me explain that.

Using your user as an example, it has the id 28:
image

Even there are 40 profiles of Chungthuang just one of that is really associated to the user 28:
image

Knowing that, what we have to do is to clean all of the profiles associated with no user?

j-peace: @chungthuang if you agree it is a good approach for that, let me know the better way to do it.
Using something like a database query or using strapi admin.

chungthuang: Yeah we should delete the ones without users. And we should have a constraint on the profile, so it's always linked to a user.

chungthuang (approved):

chungthuang:
@j-peace can you update the commit message to follow semantic release? Something like `fix(profile): Adding publishedAt to create profile with state published`.

j-peace:
Hello team.
Now we need to clean the old profiles that doesn't have realationship with users.
1) In the staging enviroment
2) In the production enviroment

j-peace:
## Cleanup status

### Staging
Before:
image

After:
image

### Production
Before:
image

After:
image

PR #295, opened 2025-08-09
Image service  - pexels integration and search by keyword

ravindrasinghshah:
TBD...

### DIFF
--- a/src/api/image/content-types/image/schema.json
+++ b/src/api/image/content-types/image/schema.json
@@ -1,0 +1,39 @@
+{
+  "kind": "collectionType",
+  "collectionName": "images",
+  "info": {
+    "singularName": "image",
+    "pluralName": "images",
+    "displayName": "Image"
+  },
+  "options": {
+    "draftAndPublish": true
+  },
+  "pluginOptions": {},
+  "attributes": {
+    "original_url": {
+      "type": "string"
+    },
+    "large_url": {
+      "type": "string"
+    },
+    "medium_url": {
+      "type": "string"
+    },
+    "small_url": {
+      "type": "string"
+    },
+    "photographer": {
+      "type": "string"
+    },
+    "photographer_url": {
+      "type": "text"
+    },
+    "image_keyword_mappings": {
+      "type": "relation",
+      "relation": "oneToMany",
+      "target": "api::image-keyword-mapping.image-keyword-mapping",
+      "mappedBy": "image"
+    }
+  }
+}
--- a/src/api/image/controllers/image.js
+++ b/src/api/image/controllers/image.js
@@ -1,0 +1,118 @@
+"use strict";
+
+/**
+ * image controller
+ */
+
+const { createCoreController } = require("@strapi/strapi").factories;
+
+module.exports = createCoreController("api::image.image", ({ strapi }) => ({
+  /**
+   * Fetch images from Pexels API and save to database
+   * Triggered on-demand for seeding images
+   * @param {Object} ctx
+   */
+  async fetchFromPexels(ctx) {
+    try {
+      const { perPage = 12, page = 1 } = ctx.query;
+
+      // Validate parameters
+      const validatedPerPage = Math.min(Math.max(parseInt(perPage), 1), 80);
+      const validatedPage = Math.max(parseInt(page), 1);
+
+      const result = await strapi
+        .service("api::image.image")
+        .fetchAndSaveImages(validatedPerPage, validatedPage);
+
+      ctx.body = {
+        success: true,
+        message: `Successfully fetched ${result.fetched} images and saved ${result.saved} new images`,
+        data: result,
+      };
+    } catch (error) {
+      ctx.status = 500;
+      ctx.body = {
+        success: false,
+        message: "Error fetching images from Pexels",
+        error: error.message,
+      };
+    }
+  },
+
+    /**
+   * Get images by keyword from image table through mapping
+   * Re-usable function
+   * @param {Object} ctx
+   * @returns {Object} Mapping Images object
+   */
+  async getImageMappings(keyword, start, limit) {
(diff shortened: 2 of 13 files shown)


src/api/image/controllers/image.js
```diff
@@ -0,0 +1,85 @@
...
+
+module.exports = createCoreController("api::image.image", ({ strapi }) => ({
+  /**
+   * Fetch images from Pexels API and save to database
+   * Triggered on-demand for seeding images
+   * @param {Object} ctx
+   */
+  async fetchFromPexels(ctx) {
+    try {
+      const { perPage = 24, page = 1 } = ctx.request.body;
```
chungthuang, line 17: How did you decide to have 24 items per page?

src/api/image/controllers/image.js
```diff
@@ -0,0 +1,85 @@
...
+
+      // Validate parameters
+      const validatedPerPage = Math.min(Math.max(parseInt(perPage), 1), 80);
```
chungthuang, line 20: The input seems to be hard coded to 24 per page though.

src/api/image/controllers/image.js
```diff
@@ -0,0 +1,85 @@
...
+   * Get images by keyword from image table through mapping
+   * Called from frontend apps
+   * @param {Object} ctx
+   */
+  async getByKeyword(ctx) {
+    try {
+      const { keyword } = ctx.params;
+      const { start = 1, limit = 12 } = ctx.request.query;
+      if (!keyword) {
+        ctx.status = 404;
```
chungthuang, line 52: This seems more like a 400 bad request than 404 not found.

workload/deployment.yaml
```diff
@@ -175,6 +175,16 @@ spec:
             secretKeyRef:
               name: v-match
               key: preferences
+        - name: PEXELS_API_URL
+          valueFrom:
+            secretKeyRef:
```
chungthuang, line 180: Does the URL need to be a secret?

src/api/image/content-types/image/schema.json
```diff
@@ -0,0 +1,45 @@
...
+      "type": "string"
+    },
+    "photographer": {
+      "type": "string"
+    },
+    "photographer_url": {
+      "type": "text"
+    },
+    "provider": {
+      "type": "text"
```
j-peace, line 36: Hi @ravindrasinghshah what is the difference between type "text" and "string"?
In this case it should be string?

src/api/image/controllers/image.js
j-peace: nice thought!

src/api/image/controllers/image.js
ravindrasinghshah: Generally 12 is a great choice for showing an image grid with pagination because it is a LCM of 2, 3 and 4. But to give a user more choices to select from I bumped the number to second LCM i.e. 24. We can then generate a image grid as e.g. small devices 2x12 grid, medium 3x8 and large 4x6.

src/api/image/content-types/image/schema.json
ravindrasinghshah: Thanks for catching this, I have changed from text to string type. `text` is used for rich text (long text) fields - which is not needed for these two fields.

workload/deployment.yaml
ravindrasinghshah: No, added pexels_api_url in configMap of staging and production. Thank you for catching.

src/api/image/controllers/image.js
chungthuang: Can we document this rationale in a comment?

src/api/image/routes/custom-image.js
```diff
@@ -0,0 +1,24 @@
+module.exports = {
+  routes: [
+    // Endpoint 1: Fetch images from Pexels and save to database (on-demand seeding)
+    {
+      method: 'POST',
```
chungthuang, line 5: Who will have permission to call this endpoint?

src/api/image/services/image.js
```diff
@@ -0,0 +1,323 @@
...
+  /**
+   * Fetch images from API based on keywords from interest table
+   * @param {number} perPage - Number of images per page (default: 12, max: 80)
+   * @param {number} page - Page number (default: 1)
+   * @returns {Promise<Array>} Array of image objects
+   */
+  async fetchImagesFromPexels(perPage = 12, page = 1) {
+    try {
+      // Get existing mapped keywords first
+      const existingMappings = await strapi.entityService.findMany(
```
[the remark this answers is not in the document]
ravindrasinghshah: For mvp I thought its enough to filter based on keyword as there's a follow-up ticket to invoke this endpoint either on cron job or trigger entry in interest table. I was thinking to change this implementation to use cursor keyword from mapping table and then select interests created after that, but I have updated this logic in this PR only.
Let me know if it answers your question. Thank you for pointing it.

dbradham:
@ravindrasinghshah do we have a link to the follow up ticket here?

src/api/image/controllers/image.js
```diff
@@ -0,0 +1,85 @@
...
+
+const { createCoreController } = require("@strapi/strapi").factories;
+
+module.exports = createCoreController("api::image.image", ({ strapi }) => ({
+  /**
+   * Fetch images from Pexels API and save to database
+   * Triggered on-demand for seeding images
+   * @param {Object} ctx
+   */
+  async fetchFromPexels(ctx) {
```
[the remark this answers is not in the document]
ravindrasinghshah: I think we discussed this on the call but still posting here for visibility to others.
No, this endpoint will be replaced eventually, once follow-up ticket to automate the image seed upon interest entry is done: https://github.com/dev-launchers/dev-launchers-platform/issues/2650

PR #301, opened 2025-10-13
feat(idea image): Add ideaImage and interest entry lifecycle event

ravindrasinghshah:
Purpose
Source images from Pexels when new entry is created in interests table.
Use ideaImage in frontend.

Changes
1. Added ideaImage field in IdeaCard table.
2. Added lifecycle event in the interests schema to call pexels for image sourcing when new entry is published or existing is updated.
3. Refactored image service to re-use pexels integration for lifecycle event and api call.

https://github.com/dev-launchers/dev-launchers-platform/issues/2650

### DIFF
--- a/src/api/image/services/image.js
+++ b/src/api/image/services/image.js
@@ -46,7 +46,7 @@ module.exports = createCoreService("api::image.image", ({ strapi }) => ({
 
     const data = await response.json();
 
-    if (data.photos && data.photos?.length > 0) {
+    if (data.photos?.length > 0) {
       // Transform response data to match schema
       const transformedImages = data.photos.map((photo) => ({
         photographer: photo.photographer,
@@ -61,7 +61,8 @@ module.exports = createCoreService("api::image.image", ({ strapi }) => ({
       }));
       return transformedImages;
     } else {
-      throw new Error(`No images found for keyword: ${keyword}`);
+      strapi.log.info(`No images found for keyword: ${keyword}`);
+      return [];
     }
   },
   /**
@@ -158,6 +159,7 @@ module.exports = createCoreService("api::image.image", ({ strapi }) => ({
       return allImages;
     } catch (error) {
       strapi.log.error("Error in fetchImagesFromPexels:", error);
+      strapi.plugin("sentry").service("sentry").sendError(error);
       throw error;
     }
   },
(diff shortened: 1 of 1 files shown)


src/api/image/services/image.js
```diff
@@ -10,6 +10,61 @@ const IMAGE_PROVIDER = {
 };
 
 module.exports = createCoreService("api::image.image", ({ strapi }) => ({
+  /**
+   * Invoke Pexels API
+   * @param {string} keyword - Keyword
```
chungthuang: Thanks for the documentation!

src/api/image/services/image.js
```diff
@@ -147,6 +163,40 @@ module.exports = createCoreService("api::image.image", ({ strapi }) => ({
...
+      //== check if keyword has images
+      const keyword = interest.interest.toLowerCase();
+      const imageCount = await this.imageCountByKeyword(keyword);
+      if (imageCount > 0) {
+        strapi.log.info(
+          `Interest ${interestId} already has ${imageCount} images`
+        );
+        return;
+      }
+      strapi.log.info(
```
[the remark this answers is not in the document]
chungthuang: Discussed during collaboration session that it's okay to have interests without images. The primary use case of images is to help users pick an image for an idea.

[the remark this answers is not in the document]
chungthuang: We also discussed that we should consider sending this message to Sentry.

src/api/image/services/image.js
```diff
@@ -147,6 +162,45 @@ module.exports = createCoreService("api::image.image", ({ strapi }) => ({
...
+      if (transformedImages.length > 0) {
+        const savedImages = await this.saveImagesToDatabase(transformedImages);
+        strapi.log.info(
+          `Saved ${savedImages.length} images to database for interest ${interestId}`
+        );
+      } else {
+        strapi.log.info(`No images found for interest: ${interestId}`);
+      }
+    } catch (error) {
+      strapi.log.error("Error in fetchAndSaveImagesForInterest:", error);
```
ravindrasinghshah: @chungthuang - I am assuming Sentry will capture `strapi.log.error` call and log it, is my understanding correct?

chungthuang: According to https://strapi.io/integrations/sentry, we need to call `strapi.plugin('sentry').service('sentry').sendError(error)`.

Issue #292, opened 2025-07-06
Member Import to Database


dbradham:

oriyomibadmus:
**Description**:
We need to import `member` data from an Excel file into the existing `members` collection in Strapi. The Excel file will not be committed to the codebase, but instead temporarily uploaded to Google Drive. The import script will `fetch` the file from Drive, process the data, and `create` entries in the collection. After a successful import, `access` to the Google Drive file will be revoked.

**Acceptance Criteria**:
 A script that can:
- [ ] Download an Google Excel file from a public Google Drive link
- [ ] Parse its contents and map to the members collection fields
- [ ] Insert each member via Strapi’s programmatic API
- [ ]  No sensitive or source data is committed to the Git repository.
- [ ]  After a successful import, the Google Drive link is disabled (or the file is deleted).
- [ ]  Imported records appear correctly in the Strapi Admin Panel under members.

**Implementation Plan**:
1. Prepare Google Drive File
2. Script Development
3. Environment Safety
4. Execution
5. Disable or delete the file from Google Drive.
6. Optionally delete the local copy of the Excel file from the server.

dbradham:
This looks good @oriyomibadmus, but we can likely reduce the complexity for an MVP by removing the requirements for Google Drive integration. This is something we will not need to frequently import in mass, so if we can just develop a script to upload the data to Strapi, from a local copy of the spreadsheet file, I think we will be good here.

dbradham:
Let's also make sure the script is committed to our backend repo under a utility folder, so we can reuse the logic to import data from a spreadsheet in the future. @oriyomibadmus

PR #308, opened 2025-11-16
Feature/member import script

godwiniheuwa:
Member Import Script (Task #292)

This PR introduces a new Member Import Utility for importing member data into the Strapi members collection from both local Excel files and Google Drive spreadsheets.

### DIFF
--- a/my_import
+++ b/my_import
@@ -1,0 +1,44 @@
+// Excel header -> Strapi attribute (api::member.member)
+const FIELD_MAPPING = {
+  // Names + contact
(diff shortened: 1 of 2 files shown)


utility/member-import/import-members-drive.mjs
dbradham: Do we need this file? It looks like a lot of similar code/functionality as `import-members-local.mjs`

godwiniheuwa: Not really we don't, it was the code I used to test locally.

PR #321, opened 2026-04-14
add github secret to deployment

oriyomibadmus:
The GitHub App credentials (GITHUB_APP_ID, GITHUB_APP_INSTALLATION_ID, GITHUB_APP_PRIVATE_KEY) are defined in the integrations secret in k8s-infra, but were never mapped to environment variables in the deployment spec. As a result, these values were undefined at runtime, causing the GitHub org invite automation to fail silently.

### DIFF
--- a/workload/deployment.yaml
+++ b/workload/deployment.yaml
@@ -165,6 +165,21 @@ spec:
             secretKeyRef:
               name: integrations
               key: sentry-dsn
+        - name: GITHUB_APP_ID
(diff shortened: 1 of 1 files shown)


chungthuang (approved):

dbradham:
@oriyomibadmus do you know why the build is failing here?

cc @chungthuang

chungthuang (approved):

PR #331, opened 2026-06-06
fix: 🐛 implemented multi case scenario for partial index

royshunhung:
Allow both camel case and snake case column name for idea_name
strapi may convert column name from camel case to snake case.
This is to addressed missing column name for variable creation

✅ Closes: #2785


royshunhung:
Closing to resolve conflicts

PR #334, opened 2026-06-07
Revert "fix: 🐛 Updated Column name for unique index"

royshunhung:
Reverts dev-launchers/strapiv4#330


src/migrators/ensure-partial-index.js
```diff
@@ -5,7 +5,7 @@ async function ensureIdeaNameActiveIndex(strapi) {
 
   await knex.raw(`
```
chungthuang, line 6: We need more than this. The while migration needs to be reverted.

